PR #64, opened 2026-07-10
Model 23847: Skill for configuring DataRobot LLM Gateway

akinhande:

## Summary
This skill should help users to configure LLM Gateway in the starter template agentic application without `dr dotenv setup`, or `dr start`. 

## Rationale
We just target LLM Gateway specific parts in the `dotenv setup` in this skill. Coding agent ask users about their preferences as if they are running the interactive CLI for configuring LLM Gateway. Coding agent never updates the `.env` file directly. It gets users preferences and creates a temporary file to be synced with `.env` variables via script. Also, users will not share their API keys with the coding agent.

---

> [!NOTE]
> **Medium Risk**
> The sync script rewrites `.env` LLM variables and can pull third-party API keys from local files; mistakes could misconfigure deployments or leak secrets if agents ignore the SKILL guardrails.
> 
> **Overview**
> **Release 1.4.0** bumps shared plugin versions and registers the new **`datarobot-llm-gateway`** skill in Gemini, the AI catalog, `CHANGELOG`, and `CODEOWNERS` (`@akinhande`).
> 
> The skill guides coding agents through LLM setup for starter agent apps **without** full `dr dotenv setup` / `dr start`: a mandatory integration choice (`gateway`, `deployed`, `external`, `blueprint-gateway`), mode-specific prompts, and strict rules so agents never paste secrets or hand-edit `.env`.
> 
> **`list_gateway_models.py`** calls the LLM Gateway catalog API using credentials from `.env`/env and returns active models as JSON (no token in output). **`sync_llm_env.py`** writes external-provider templates to `.secrets/llm-external.env`, merges `.datarobot/llm-config.json` into `.env` (stripping prior LLM-managed keys), validates deployment IDs, and optionally deletes the interim config after a successful sync.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 624868cd354ebd86bff1cd438fe79b94d9cdee78. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/CODEOWNERS
```diff
@@ -15,6 +15,9 @@
 /skills/datarobot-model-training/ @datarobot/core-modeling
 /skills/datarobot-predictions/ @datarobot/core-modeling
 
+# Skills owned by Predictive AI team
+/skills/datarobot-llm-gateway/ @akinhande
```
carsongee: Is there a team we can use as opposed to your individual account? We prefer that for continuity. Perhaps @datarobot/core-modeling ?

skills/datarobot-llm-gateway/SKILL.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+Tell the user:
+
+> Open `.secrets/llm-external.env`, fill in the values, save, then reply **credentials ready**.
+> Do not paste secrets in chat.
+
+Wait for **credentials ready** before Step 5.
+
+---
```
carsongee: This should suggest using our existing credentials store `($XDG_CONFIG|~/.config)/datarobot/drconfig.yaml` rather than inventing a new one and spreading keys

skills/datarobot-llm-gateway/SKILL.md
```diff
@@ -0,0 +1,315 @@
...
+
+## Step 3 — Credentials file plan (external only)
+
+Show the user this plan before creating files:
+
+```text
+.secrets/llm-external.env   ← you fill provider secrets here (gitignored)
+.datarobot/llm-config.json  ← non-secret LLM settings (gitignored)
```
carsongee: We should always recommend `$XDG_CONFIG or ~/.config/datrobot` for any additional configurations required for our platform

skills/datarobot-llm-gateway/SKILL.md
```diff
@@ -0,0 +1,315 @@
...
+5. Only `sync_llm_env.py` merges LLM keys into `.env` — do not edit `.env` manually
+6. Run all commands from **project root**
+7. After sync: `dr dotenv validate` then remind user: `dr task run infra:up-yes`
+
+---
+
+## Step 0 — Prerequisites
+
+1. Project root must exist (`.datarobot/cli/llm.yml` present)
+2. If no `.env`: tell user to run `dr dotenv setup --if-needed` or `dr start` first (base vars only)
```
carsongee: If they have this, then things are already configured, and just running `dr dotenv setup` will prompt them through to the nice LLM Selection screen in the onboarding.

skills/datarobot-llm-gateway/SKILL.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+## Hard rules
+
+1. **Never** ask the user to paste API keys or `DATAROBOT_API_TOKEN` in chat
+2. **Never** read, copy, echo, or pass `DATAROBOT_API_TOKEN` yourself. Only the
+   bundled scripts touch the token; they read it from `.env` / environment and
+   never emit it to stdout. Do not run `cat .env`, `env | grep TOKEN`,
+   `echo $DATAROBOT_API_TOKEN`, `curl -H "Authorization: Bearer $..."`, or any
+   equivalent one-liner
```
carsongee: I'd generally suggest that you prompt this to use `dr auth login` and then use the XDG_CONFIG/datarobot/drconfig.yaml` for credential access. It is a nice browser based flow to log the user and get a key for them

carsongee (changes requested):

.github/CODEOWNERS
akinhande: I changed it. I am in the predictive-ai team. :)

skills/datarobot-llm-gateway/SKILL.md
akinhande: You're right! I changed this part. Now it copies the `.env.template` when `.env` is missing. Later, the skill will sync user preferences for LLM Gateway to `.env`. Or do you know a better way maybe?

skills/datarobot-llm-gateway/scripts/list_gateway_models.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
```
carsongee: Now that we have `dr llm-gateway list` in the CLI, I don't think this is needed any longer

carsongee (changes requested):

skills/datarobot-llm-gateway/scripts/list_gateway_models.py
akinhande: Good point! Thanks for letting me know.

skills/datarobot-llm-gateway/SKILL.md
```diff
@@ -0,0 +1,338 @@
...
+   Next step (Step 3) will create that file as a blank template if it
+   doesn't exist. Please fill it in your own editor — do not paste the
+   values in chat — then tell me "credentials ready" and I'll re-run the
+   sync.
+   ```
+
+   Do not create the file yourself, do not `cat` it, and do not accept
+   secret values in chat.
+
+---
```
carsongee: This section duplicates the logic encoded in the AF component:
https://github.com/datarobot-community/af-component-llm/blob/main/template/.datarobot/cli/llm.yml.jinja#L107-L214

That the CLI wizards the user through. I'm concerned that will create mismatch and maintenance issues. How about instead, we instruct the LLM here to read the `.datarobot/cli/llm.yml` file to discover the supported items and credentials. 

As a specific example, we recently added support for "OpenAI Compatible" here, and this PR is missing that option:
https://github.com/datarobot-community/af-component-llm/blob/bbe936f355498b8d8e7e128f0146a8eb02a155a5/template/.datarobot/cli/llm.yml.jinja#L156-L163

skills/datarobot-agent-llm-selection/scripts/sync_llm_env.py
```diff
@@ -0,0 +1,221 @@
...
+        out.append(f"{key}={_quote(llm_vars[key])}")
+    env_path.write_text("\n".join(out) + "\n", encoding="utf-8")
+
+    print(f"Synced {len(llm_vars)} LLM variable(s) into {env_path}")
+    for key in sorted(llm_vars):
+        print(f"  ✓ {key}")
+
+    print(
+        "\nNext (run in your terminal — these echo secrets):\n"
+        "  dr dotenv validate\n"
```
carsongee, line 256: We should run this here and exit with a bad error code if it isn't valid along with the stdout/stderr from the command. It would make this less error prone for the Agent/LLM to operate

skills/datarobot-llm-gateway/SKILL.md
```diff
@@ -0,0 +1,338 @@
...
+   - If both `.env` and no `.env.template` are missing, tell the user they're not in
+     a DataRobot agent project root and stop.
+
+---
+
+## Step 1 — Integration mode (ASK THIS FIRST — MANDATORY)
+
+**Before doing anything else in this skill**, ask the user which integration
+mode they want. The value must be one of exactly these four:
+`gateway`, `deployed`, `external`, `blueprint-gateway`.
```
carsongee: I'd prefer we stick to our naming in the LLM config: https://github.com/datarobot-community/af-component-llm/blob/bbe936f355498b8d8e7e128f0146a8eb02a155a5/template/.datarobot/cli/llm.yml.jinja#L29-L47

It also includes additional options than are presented here. This raises another maintenance issue.

Perhaps we could instead teach it here to reference the options in the CLI config and leave it to that. i.e., `read .llm.yml for the `env: INFRA_ENABLE_LLM`` key and all the help string and options there.

Just spitballing here, but maybe a simple script to pull out the help text and options directly? i.e., `get_infra_llm_help.py` or something more generic in the `scripts` folder that can rifle through the yaml and present the user with the real options in their live yaml file? That would solve the versioning and maintenance problem for this and the credentials work

carsongee (changes requested):
We're not quite to where this skill will work with any version of the template, but it looks like we're getting very close. Thank you for your time and patience with us!

skills/datarobot-agent-llm-selection/scripts/sync_llm_env.py
akinhande: Good catch! The problem is `dr dotenv validate` prints the whole `.env` file, so `DATAROBOT_API_TOKEN` is in its stdout. 
So, this is a work around that by running it as `dr dotenv validate >/dev/null 2>&1` and reading only the exit code; the agent never sees the output, and on failure it tells the user to run the command themselves in their own terminal.

skills/datarobot-llm-gateway/SKILL.md
akinhande: Good point! I didnt know about this file. I implemented the suggested changes.

skills/datarobot-llm-gateway/SKILL.md
akinhande: Thanks for the suggestions. I like the idea. I implemented the changes.

carsongee (approved):

carsongee:
Thank you for the contribution!

I made a few minor changes at the end:
* Added a backup function prior to changing the .env
* Renamed it to `datarobot-agent-llm-selection` for better context and discovery
* Added support for agent-assist spec intelligence for recommending LLMs

PR #67, opened 2026-07-16
[APP-6555] Add Swarm simulation

elvinagam:
## Summary

- Adds `datarobot-agent-assist-simulate` as a new top-level skill for adversarially testing and hardening AI agents before deployment
- Native harness-orchestrated architecture: no DataRobot credentials required, runs entirely via `dr opencode serve`
- Includes scenario generation (attack, behavior, persistence tracks), parallel swarm execution, convergence loop with automated prompt patching, selective e2e tool execution, and full evaluation reporting

## What's included

- `SKILL.md` — full harness instructions
- 8 role prompts (generators, runner, fixture, evaluator, fixer, diagnoser)
- 11 Python scripts (native execution state machine, swarm orchestration, convergence loop, gateway worker, tool executor, report writer)
- 7 integration test files (~3000 lines)
- CODEOWNERS + ai-catalog entries

## Notes

- The restructure of `datarobot-agent-assist/` into `agent-assist-build/` sub-skill is a follow-up PR
- Full history preserved on `backup/swarm-opencode-full` branch

---

> [!NOTE]
> **Medium Risk**
> Large new orchestration surface that can mutate `agent_spec.md` and implementation prompts during convergence; selective e2e can invoke real readonly tools from the user's `tools.py`.
> 
> **Overview**
> Introduces **adversarial swarm simulation** for implemented agents and turns the top-level `datarobot-agent-assist` skill into a thin router with a new **Battle-test** menu option.
> 
> **`agent-assist-simulate`** documents a five-step harness workflow (configure → generate scenarios → parallel swarm run → convergence with prompt/code fixes → `eval_report.md`). It drives **`dr opencode serve`** and Python orchestrators: scenario prep/validation, per-scenario runner/fixture/evaluator state machine, parallel **`native_swarm.py`**, convergence **`advance`/`report`**, optional **selective e2e** readonly tool execution via **`tool_executor.py`**, and LLM workers through **`gateway_worker.py`**. Role prompts define strict JSON outputs for attack/behavior/persistence generators, simulation runner, fixtures, and evaluation.
> 
> **`agent-assist-build`** holds the former monolithic design/code/deploy/dress-rehearsal content and routes post-coding users to simulate. Root **`SKILL.md`** is reduced to a four-option menu and explicit “read sub-skill” rules.
> 
> Adds **`pydantic>=2.0`** to integration test deps plus integration tests for gateway worker and native convergence (and related smoke config).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eeca00a2ad511ae9112fe3d7b67219ab89ef16f1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


skills/datarobot-agent-assist/agent-assist-simulate/scripts/tool_executor.py
```diff
@@ -0,0 +1,78 @@
...
+    if not tools_path.is_file():
+        print(f"tools-path not found: {tools_path}", file=sys.stderr)
+        sys.exit(1)
+    spec = importlib.util.spec_from_file_location("_user_tools", tools_path)
+    if spec is None or spec.loader is None:
+        print(f"cannot load module from: {tools_path}", file=sys.stderr)
+        sys.exit(1)
+    module = importlib.util.module_from_spec(spec)
+    try:
+        spec.loader.exec_module(module)
```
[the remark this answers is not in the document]
elvinagam: :done

skills/datarobot-agent-assist/agent-assist-simulate/SKILL.md
```diff
@@ -0,0 +1,369 @@
...
+
+---
+
+## Step 3 — Simulate
+
+If `.datarobot/swarm/` exists, ask the user:
+> "Previous results exist — run again and archive them? (yes/no)"
+
+If yes, archive the directory with a timestamp (`mv .datarobot/swarm .datarobot/swarm-<timestamp>`)
+and continue. If no, stop and let the user review previous results.
```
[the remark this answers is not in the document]
elvinagam: adressed

skills/datarobot-agent-assist/agent-assist-simulate/SKILL.md
```diff
@@ -0,0 +1,369 @@
...
+
+   where N is the current turn number and M is the scenario's `max_turns`. Read N and the next role
+   from each `submit` response (`turn_number`, `role`).
+
+   ```bash
+   <python> <skill_scripts_dir>/gateway_worker.py \
+     --role-prompt <run-scenario|generate-tool-return|evaluate-result> \
+     --input-path <input_path> \
+     --response-path <response_path> \
+     --model <model> --server-url <opencode_server_url>
```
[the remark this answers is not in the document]
elvinagam: added

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_swarm.py
```diff
@@ -0,0 +1,752 @@
...
+        str(input_path),
+        "--response-path",
+        str(response_path),
+        "--tools-path",
+        str(tools_path),
+    ]
+    result = subprocess.run(cmd, capture_output=True, text=True)
+    if result.returncode != 0:
+        print(result.stderr or result.stdout, file=sys.stderr)
+    return result.returncode == 0
```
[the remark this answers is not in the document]
elvinagam: fixed

skills/datarobot-agent-assist/agent-assist-simulate/SKILL.md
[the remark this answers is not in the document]
elvinagam: done

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_swarm.py
```diff
@@ -0,0 +1,752 @@
...
+            "fail",
+            "--run-dir",
+            str(run_dir),
+            "--reason",
+            reason,
+        ],
+        capture_output=True,
+        text=True,
+    )
+
```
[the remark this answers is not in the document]
elvinagam: addressed

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_convergence.py
```diff
@@ -0,0 +1,368 @@
...
+    _resolve_project_file,
+    _resolve_under_root,
+    _scenario_id,
+    load_criteria,
+    load_json,
+    load_native_config,
+    load_spec,
+    write_json,
+)
+from contracts import (
```
carsongee: Shouldn't we make a python package so we can do `from .contracts` here? If you happen to have https://pypi.org/project/contracts/ installed for example, it will break

skills/datarobot-agent-assist/agent-assist-simulate/SKILL.md
```diff
@@ -0,0 +1,409 @@
...
+
+```bash
+<python> -c "import sys; assert sys.version_info >= (3, 11)"
+<python> -c "import pydantic, yaml" 2>/dev/null || <python> -m pip install pydantic pyyaml
+```
+
+**OpenCode:**
+
+```bash
+opencode --version 2>/dev/null || (dr opencode upgrade && export PATH="$PATH:$HOME/.opencode/bin")
```
carsongee: Don't you want to force `dr opencode` ? It ensure the skills are installed, whereas `opencode` may not hav eyour skills

carsongee (commented):
There's enough code in here that it feels like a library or cli as much as an LLM operated skill. I'd consider making a new library like datarobot-agent-assist so that you can have it pip install it, and then execute it from there.

Overall, the structure and skills look good

skills/datarobot-agent-assist/agent-assist-simulate/scripts/gateway_worker.py
```diff
@@ -0,0 +1,290 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026 DataRobot, Inc. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Thin LLMGateway worker adapter.
+
+Invokes `dr opencode run` in an isolated temporary directory, parses the
```
victorborshak, line 7: @elvinagam This seems to be very opecode focused. Does it mean it's not going to work in the other coding agents like claude, cursor etc? IMO we should aim the skill to be fully functional in all coding agents.

victorborshak:
@elvinagam What's the reason the `scripts/check_codespace.py` wasn't moved to the `agent-assist-build/scripts` directory?

I tried to run in dr opencode and swarm simulation worked, but the main Agent Assist skill did not offer me to run it automatically, I had to ask specifically about swarm simulation. This makes it hard for the users to discover that swarm simulation functionality exists.


Screenshot 2026-07-23 at 15 10 18
Screenshot 2026-07-23 at 15 13 10

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_convergence.py
elvinagam: you're right it's fragile. I'll rename contracts.py to a unique name (dr_swarm_contracts.py or similar) . is that good for now?

skills/datarobot-agent-assist/agent-assist-simulate/SKILL.md
elvinagam: absolutely

skills/datarobot-agent-assist/agent-assist-simulate/scripts/gateway_worker.py
elvinagam: The SKILL.md works in any coding agent -  swarm is a subprocess the agent shells out to, not the agent interface. It uses dr opencode run as the LLM backend to route through LLM Gateway to drive tokens

elvinagam:
> @elvinagam What's the reason the `scripts/check_codespace.py` wasn't moved to the `agent-assist-build/scripts` directory?
> 
> I tried to run in dr opencode and swarm simulation worked, but the main Agent Assist skill did not offer me to run it automatically, I had to ask specifically about swarm simulation. This makes it hard for the users to discover that swarm simulation functionality exists.
> 
> Screenshot 2026-07-23 at 15 10 18 Screenshot 2026-07-23 at 15 13 10

check_codespace.py is at the parent level. i thought the parent router SKILL.md references it as a shared pre-flight step across sub-skills, not specific to build. is that correct?

Good catch on discoverability. I tightened the activation behavior

skills/datarobot-agent-assist/agent-assist-simulate/scripts/tool_executor.py
```diff
@@ -0,0 +1,81 @@
...
+    tool_name: str = package.get("tool_name", "")
+    call_args: dict[str, object] = package.get("args", {})
+
+    if not tool_name:
+        print("input package missing tool.function_name", file=sys.stderr)
+        sys.exit(1)
+
+    module = _load_tools_module(args.tools_path)
+
+    fn = getattr(module, tool_name, None)
```
chasdr, line 83: hey, Elvin 🙂 one thing on the real-execution path i'm not sure about (i may be missing context):

native_swarm gates tool_executor to is_readonly tools via e2e_tools, but tool_executor itself will call any function name in the input package. and the convergence rerun path looks like it leans on the orchestrating agent to pick tool_executor vs gateway_worker rather than a code gate.

since the whole point is feeding adversarial scenarios, would it be worth passing the readonly set into tool_executor and refusing anything not marked is_readonly? that way the "only readonly runs for real" guarantee holds even if the upstream selection is wrong.

skills/datarobot-agent-assist/agent-assist-simulate/scripts/gateway_worker.py
```diff
@@ -0,0 +1,290 @@
...
+# One project-local log for swarm and convergence workers (relative to agent cwd).
+METRICS_PATH = Path(".datarobot/swarm/metrics.jsonl")
+PROMPT_ROLE_MAP = {
+    "generate-attack.md": "generator/attack",
+    "generate-behavior.md": "generator/behavior",
+    "generate-persistence.md": "generator/persistence",
+    "run-scenario.md": "runner",
+    "generate-tool-return.md": "fixture",
+    "evaluate-result.md": "evaluator",
+    "generate-fix.md": "fixer",
```
chasdr: PROMPT_ROLE_MAP has generate-fix.md and diagnose-failure.md, but those prompt files aren't in the PR and i don't see anything invoking the fixer/diagnoser roles. the description mentions 8 prompts (fixer, diagnoser) but 6 shipped.

are those coming in a follow-up?

skills/datarobot-agent-assist/agent-assist-simulate/scripts/tool_executor.py
```diff
@@ -0,0 +1,81 @@
...
+        return_value = fn(**call_args)
+    except Exception as exc:
+        print(f"{tool_name} raised {type(exc).__name__}: {exc}", file=sys.stderr)
+        sys.exit(1)
+
+    response = {"tool_name": tool_name, "args": call_args, "return_value": return_value}
+    args.response_path.parent.mkdir(parents=True, exist_ok=True)
+    tmp = args.response_path.with_name(f".{args.response_path.name}.tmp")
+    tmp.write_text(json.dumps(response, ensure_ascii=False, indent=2), encoding="utf-8")
+    os.replace(tmp, args.response_path)
```
chasdr, line 111: the try/except only wraps the fn() call, so if a real tool returns something non-JSON-serializable the json.dumps below throws an uncaught traceback instead of the clean exit(1) you use everywhere else. small suggestion:

```suggestion
    response = {"tool_name": tool_name, "args": call_args, "return_value": return_value}
    try:
        payload = json.dumps(response, ensure_ascii=False, indent=2)
    except TypeError as exc:
        print(f"{tool_name} return value is not JSON-serializable: {exc}", file=sys.stderr)
        sys.exit(1)
    args.response_path.parent.mkdir(parents=True, exist_ok=True)
    tmp = args.response_path.with_name(f".{args.response_path.name}.tmp")
    tmp.write_text(payload, encoding="utf-8")
    os.replace(tmp, args.response_path)
```

skills/datarobot-agent-assist/agent-assist-simulate/scripts/tool_executor.py
elvinagam: I agree. thanks for the review, yes, fixed now.

skills/datarobot-agent-assist/agent-assist-simulate/scripts/gateway_worker.py
elvinagam: opps. these was prior prompts we used before I simplified it. Cleaned up those now.

skills/datarobot-agent-assist/agent-assist-simulate/scripts/tool_executor.py
elvinagam: sure, thanks.

skills/datarobot-agent-assist/agent-assist-simulate/SKILL.md
```diff
@@ -0,0 +1,409 @@
...
+
+   ```bash
+   # 1. Call worker using input_path and response_path from initialize (or previous submit).
+   #    For the fixture role: if agent_config.yaml has execution.mode: selective_e2e and
+   #    tools.py exists, call tool_executor.py instead of gateway_worker.py:
+   #
+   #    selective_e2e fixture:
+   #    <python> <skill_scripts_dir>/tool_executor.py \
+   #      --input-path <input_path> --response-path <response_path> \
+   #      --tools-path <tools_path>
```
chasdr: the changes you made look good 👍 but i think we have a doc/code mismatch:

--readonly-tools is required=True now, and native_swarm passes it, but this convergence invocation still shows just --input-path/--response-path/--tools-path. an agent following this path hits `error: the following arguments are required: --readonly-tools` (i reproduced it).

mind adding --readonly-tools here too?

```suggestion
   #    <python> <skill_scripts_dir>/tool_executor.py \
   #      --input-path <input_path> --response-path <response_path> \
   #      --tools-path <tools_path> --readonly-tools <comma-separated readonly fn names>
```

chasdr (approved):
lgtm 👍 only open thing is the tool_executor.py `--readonly-tools` mismatch in the SKILL.md convergence block i flagged above. not a merge blocker from my side.

adding @victorborshak for a second set of eyes before this goes in.

skills/datarobot-agent-assist/agent-assist-simulate/SKILL.md
elvinagam: absolutely

victorborshak:
> check_codespace.py is at the parent level. i thought the parent router SKILL.md references it as a shared pre-flight step across sub-skills, not specific to build. is that correct?
> 
> Good catch on discoverability. I tightened the activation behavior

@elvinagam I would recommend to consider moving `check_codespace.py` (and the related instruction too) into the `agent-assist-build` skill because
- the logic in the `check_codespace.py` is actually related to the agent building skill - it checks if ports are exposed in the codespace
- `check_codespace.py` references the `env_utils.py` so as of now they need to be in the same directory to work properly
- There's no definition of `<skill_scripts_dir>` which is used in the root `SKILL.md` right now
- the whole directory/file structure of `datarobot-agent-assist` would look much cleaner this way

Also, I would strongly advice to do some testing of the updated skill in the Codespace to make sure everything is working as expected.

victorborshak (commented):
@elvinagam  I left some more comments about `check_codespace.py` instructions and about testing in the Codespace.

elvinagam:
> > check_codespace.py is at the parent level. i thought the parent router SKILL.md references it as a shared pre-flight step across sub-skills, not specific to build. is that correct?
> > Good catch on discoverability. I tightened the activation behavior
> 
> @elvinagam I would recommend to consider moving `check_codespace.py` (and the related instruction too) into the `agent-assist-build` skill because
> 
> * the logic in the `check_codespace.py` is actually related to the agent building skill - it checks if ports are exposed in the codespace
> * `check_codespace.py` references the `env_utils.py` so as of now they need to be in the same directory to work properly
> * There's no definition of `<skill_scripts_dir>` which is used in the root `SKILL.md` right now
> * the whole directory/file structure of `datarobot-agent-assist` would look much cleaner this way
> 
> Also, I would strongly advice to do some testing of the updated skill in the Codespace to make sure everything is working as expected.

thanks @victorborshak - made the update for codespaces script to build folder.

shreyaag-dr (commented):
lgtm

victorborshak (approved):
@elvinagam we probably need to bump the plugin version in the .json files.

skills/datarobot-agent-assist/SKILL.md
```diff
@@ -3,24 +3,25 @@ name: datarobot-agent-assist
 description: >-
   Use when the user wants to design, build, code, simulate, or deploy an AI agent (not a predictive
   model) to DataRobot; mentions agent_spec.md, dr-assist, datarobot-agent-assist, dress rehearsal,
-  or the DataRobot agent template; wants to scaffold a LangGraph, CrewAI, LlamaIndex, NAT, or Base
+  swarm simulation, or the DataRobot agent template; wants to scaffold a LangGraph, CrewAI, LlamaIndex, NAT, or Base
```
carsongee, line 6: Do you think users would think to say this? I would think "harden this agent" would be the chat we want to pick up on and run the simulation. It didn't pick this skill when I said that

elvinagam: sure, I can replace that with harden the agent or simpler wording

elvinagam: ✅

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_convergence.py
```diff
@@ -0,0 +1,368 @@
...
+        default=Path(".datarobot/swarm/convergence"),
+    )
+    initialize_parser.add_argument("--actual-model")
+
+    advance_parser = subparsers.add_parser("advance")
+    advance_parser.add_argument("spec", type=Path)
+    advance_parser.add_argument(
+        "--convergence-dir",
+        type=Path,
+        default=Path(".datarobot/swarm/convergence"),
```
carsongee: These paths are used all over the place, and they generated about 75 files. They are not added to .gitignore, so I check them in by default, but they are not useful. Please write this code to use named temporary files.

If that isn't possible due to sandboxes or similar, than have these scripts ensure that `.datarobot/swarm` is added to .gitignore, .dockerignore, .wapiignore, etc.

tests/integration/test_tool_executor.py
```diff
@@ -0,0 +1,246 @@
...
+import pytest
+import yaml
+
+REPO_ROOT = Path(__file__).parent.parent.parent
+SIMULATE_DIR = REPO_ROOT / "skills" / "datarobot-agent-assist" / "agent-assist-simulate"
+EXECUTOR = SIMULATE_DIR / "scripts" / "tool_executor.py"
+sys.path.insert(0, str(SIMULATE_DIR / "scripts"))
+artifacts = importlib.import_module("artifacts")
+contracts = importlib.import_module("swarm_contracts")
+native = importlib.import_module("native_execution")
```
carsongee, line 20: I'm not sure where to put this, but the agent broke my tool with agy and it choosing flash 2.5:
https://github.com/datarobot-oss/yourself-as-a-service/pull/7/changes#diff-74df0cc914f7063ba33bbe9445fdf3c6d0b1df12fec774f55a624f8d35bb2f90R42-R55

It also broke my "universal-truths"/global agent config that says to never add import statements outside of the top of a Python file.

skills/datarobot-agent-assist/agent-assist-simulate/scripts/gateway_worker.py
```diff
@@ -0,0 +1,288 @@
...
+        "--rejection-note",
+        default=None,
+        help="rejection reason to append to message on retry",
+    )
+    parser.add_argument(
+        "--timeout",
+        type=int,
+        default=120,
+        help="subprocess timeout in seconds (default: 120)",
+    )
```
carsongee, line 182: I hit this a bunch. It is good to have the argument so it could unstick itself, but I'd suggest a longer default is warranted. I know we want to fail fast, but running the same command multiple times with increasing timeout values makes the whole thing take a lot longer if it will eventually work anyway. I hit this with gpt-5.5 as the model

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_convergence.py
```diff
@@ -0,0 +1,368 @@
...
+    )
+
+    report_parser = subparsers.add_parser("report")
+    report_parser.add_argument("spec", type=Path)
+    report_parser.add_argument(
+        "--convergence-dir",
+        type=Path,
+        default=Path(".datarobot/swarm/convergence"),
+    )
+    report_parser.add_argument("--output", type=Path, default=Path("eval_report.md"))
```
carsongee, line 321: Same ignore question here. Are we expecting the user to want to commit this file? Seems to me we want to make a folder for artifacts we don't need to live with the code base

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_scenarios.py
```diff
@@ -0,0 +1,322 @@
...
+        "--work-dir", type=Path, default=Path(".datarobot/swarm")
+    )
+
+    confirm_parser = subparsers.add_parser("confirm")
+    confirm_parser.add_argument(
+        "--work-dir", type=Path, default=Path(".datarobot/swarm")
+    )
+    confirm_parser.add_argument(
+        "--output", type=Path, default=Path("evaluation_criteria.md")
+    )
```
carsongee, line 286: Same .gitignore/storage question here as well

carsongee (changes requested):
The .datarobot/swarm stuff and the output files are my only hard-blockers for letting this merge. I ran this with copilot CLI: https://github.com/datarobot-oss/yourself-as-a-service/pull/8

and antigravity/agy CLI: https://github.com/datarobot-oss/yourself-as-a-service/pull/7/changes#diff-5f0ab18b64508efbf63111d5b7a5392de861fb421b7125c22abb6e2a0d40e257

Agy asked no questions and did not produce a report or run the simulation.  That feels like a miss that we can work to refine after this merges.

I feel a product requirement was missed for model selection. You're fanning out and running 12 sessions. Each session could be work better performed by various of the 80+ models we have running, so I'd suggest we adjust the skill to prompt for that, but again, that come in a later PR if you want to get broader feedback.

skills/datarobot-agent-assist/agent-assist-simulate/scripts/gateway_worker.py
```diff
@@ -0,0 +1,288 @@
...
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+import time
+from pathlib import Path
+
+# One project-local log for swarm and convergence workers (relative to agent cwd).
+METRICS_PATH = Path(".datarobot/swarm/metrics.jsonl")
```
carsongee: should be a tempdir that self deletes

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_convergence.py
elvinagam: addressed

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_scenarios.py
elvinagam: addressed

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_convergence.py
elvinagam: sure

skills/datarobot-agent-assist/agent-assist-simulate/scripts/gateway_worker.py
elvinagam: would we like to bump it to higher number later after user testing?

tests/integration/test_tool_executor.py
elvinagam: interesting. is this Stochasticity of harnesses?

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_swarm.py
```diff
@@ -0,0 +1,779 @@
...
+                    tools_path,
+                )
+            ] = task
+        for future in as_completed(futures):
+            task = futures[future]
+            completed += 1
+            elapsed = time.monotonic() - t0
+            try:
+                status = future.result()
+            except Exception as exc:
```
carsongee: Optional nit: I think you could pair this down to a more narrow exception on a futures object. It can only be OSError, ValidationError, or ValueError. Those are different enough it is probably worth trying to handle some of them

skills/datarobot-agent-assist/agent-assist-simulate/scripts/artifacts.py
```diff
@@ -0,0 +1,183 @@
...
+    except yaml.YAMLError as exc:
+        raise CriteriaError(f"{path} contains invalid YAML: {exc}") from exc
+
+    if not isinstance(data, list) or not data:
+        raise CriteriaError(f"{path} must contain a non-empty list of scenarios")
+
+    try:
+        return [Scenario.model_validate(scenario) for scenario in data]
+    except Exception as exc:
+        raise CriteriaError(f"{path} contains an invalid scenario: {exc}") from exc
```
carsongee: Should just be "ValidationError". Anything else is from you, the programmer

skills/datarobot-agent-assist/agent-assist-simulate/scripts/artifacts.py
```diff
@@ -0,0 +1,183 @@
...
+                "execution": {
+                    "mode": "simulated",
+                    "requested_scope": {"tools": [], "resources": []},
+                },
+            }
+        )
+    except ConfigError:
+        raise
+    except Exception as exc:
+        raise ConfigError(f"{path} contains invalid configuration: {exc}") from exc
```
carsongee: Same here, would swallow KeyError, AttributeError, etc. Should just be ValidationError

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_execution.py
```diff
@@ -0,0 +1,447 @@
...
+
+import argparse
+import json
+import sys
+from pathlib import Path
+from typing import Literal
+
+from pydantic import Field, TypeAdapter, ValidationError
+
+from artifacts import _one_line, load_criteria, load_json, load_spec, write_json
```
carsongee: Importing the underscore breaks the whole Python convention of private _....should never import `_<name>` things. Conveniently, I don't personally find the convention useful anyway. So I'd suggest renaming it `one_line` regardless

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_scenarios.py
```diff
@@ -0,0 +1,329 @@
...
+import sys
+from pathlib import Path
+from typing import Literal
+
+from pydantic import ValidationError
+
+from artifacts import (
+    _one_line,
+    _resolve_project_file,
+    _resolve_under_root,
```
carsongee: More "private" imports. Please correct them

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_swarm.py
```diff
@@ -0,0 +1,779 @@
...
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from pathlib import Path
+
+from pydantic import ValidationError
+
+from artifacts import (
+    _one_line,
+    _resolve_project_file,
+    _resolve_under_root,
+    _scenario_id,
```
carsongee: And more private imports

skills/datarobot-agent-assist/agent-assist-simulate/scripts/gateway_worker.py
```diff
@@ -0,0 +1,290 @@
...
+    return None
+
+
+def _write_metrics(record: dict[str, object]) -> None:
+    try:
+        METRICS_PATH.parent.mkdir(parents=True, exist_ok=True)
+        with METRICS_PATH.open("a", encoding="utf-8") as f:
+            f.write(json.dumps(record, ensure_ascii=False) + "\n")
+    except OSError:
+        pass
```
carsongee, line 144: All workers writing to the same path is a concurrency issue. On Mac and Linux, this is pretty unlikely to break the program, but on Windows, where only one program can open a file at a time and this append may not be atomic, it will hit this OSError that you're just ignoring. I'd strongly suggest each worker writes their own file, and then you append them at the end. i..e. add ye-olde `uuid.uuid4().hex` to the file name. Then merge them with a helper in `artifacts.py` using a glob to grab them

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_convergence.py
```diff
@@ -0,0 +1,371 @@
...
+        default=Path(os.environ.get("SWARM_DIR", ".datarobot/swarm")) / "convergence",
+    )
+    initialize_parser.add_argument("--actual-model")
+
+    advance_parser = subparsers.add_parser("advance")
+    advance_parser.add_argument("spec", type=Path)
+    advance_parser.add_argument(
+        "--convergence-dir",
+        type=Path,
+        default=Path(os.environ.get("SWARM_DIR", ".datarobot/swarm")) / "convergence",
```
carsongee: Every single one of these used the default. The SWARM_DIR environment variable setting is not working reliably. I think you should write some code to write this out to a file if not set:
```python
if not SWARM_DIR:
   if Path.exists('.datarobot/swarm/swarm_temp'):
       return swarm_temp.read()
   temp_dir = ...
   with open('.datarobot/swarm/swarm_temp') as f:
        f.write(temp_dir)
   return temp_dir
```
as a helper function somewhere, and use it everywhere

carsongee (approved):
Once you've addressed the comments, good to merge! 🔥

skills/datarobot-agent-assist/agent-assist-simulate/scripts/gateway_worker.py
elvinagam: already addressed

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_swarm.py
elvinagam: added

skills/datarobot-agent-assist/agent-assist-simulate/scripts/artifacts.py
elvinagam: fixed

skills/datarobot-agent-assist/agent-assist-simulate/scripts/artifacts.py
elvinagam: agreed.

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_execution.py
elvinagam: ahh, these slipped through - fixed

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_scenarios.py
elvinagam: yeap yeap

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_swarm.py
elvinagam: more fixes

skills/datarobot-agent-assist/agent-assist-simulate/scripts/gateway_worker.py
elvinagam: just fixed that edge case. we were ignoring it as I thought it wouldn't be needed to interrupt swarm because of it

skills/datarobot-agent-assist/agent-assist-simulate/scripts/native_convergence.py
elvinagam: added a resolve function

PR #71, opened 2026-07-26
[CFX-6540] dr dependency check [CFX-6541] Agent subdirectory setup[CFX-7179]Enhance Resume Agent Design Logic

NolanKong:
## Summary
Change the dr dependency check: add dr dependency install.
Now AA supports specify subdirectory to setup agent
Further optimize the AA workflow and added the resume design logic
## Rationale
https://datarobot.atlassian.net/browse/CFX-6540
https://datarobot.atlassian.net/browse/CFX-6541
https://datarobot.atlassian.net/browse/CFX-7179
https://datarobot.atlassian.net/browse/CFX-6746

---

> [!NOTE]
> **Low Risk**
> Changes are confined to Agent Assist skill docs and local helper scripts; no production services or auth logic in the main product, though agents must follow the new `--target-dir` and dependency-install flow to avoid wrong `.env` paths.
> 
> **Overview**
> **`datarobot-agent-assist`** gets a large workflow and documentation refactor: **`SKILL.md`** is trimmed and gains **Workflow Discipline**, session flags (`<prerequisites_passed>`, `<workspace_resolved>`, `<design_to_code>`, `<design_messy_cwd>`), **`.env` only under `<target_dir>`**, and pointers into new/expanded references instead of inlined CLI, schema, helper scripts, and dress-rehearsal text. The old **Before Coding Begins** gate is folded into **`pre-coding-checklist.md` Bootstrap step 2** with explicit pass/fail reporting, **Spec issues** routing to **`resume-design.md`**, missing-spec recovery, and **clone discipline** (spec-only auto-clone vs messy cwd needing step 7 confirmation—**Code the agent** is not confirmation).
> 
> **Design flow** adds **`resume-design.md`** (spec-complete rules, resume table, no skip to coding), reorders **Post-design** menu (code first), and tightens transitions (dress rehearsal in its own turn). **Workspace resolution** adds **Path resolution**, asks whether cwd is project root for Code, and wires **resume design** on continue-existing-spec.
> 
> **Dependency handling** is centralized in **`dependency-validation.md`**: `dr dependency check`, then **`dr dependency install --yes`**, then re-check (pre-coding step 10 and pre-deploy step 4).
> 
[...]

PR #73, opened 2026-07-29
[CFX-6981] Source LLM model selection from the DataRobot CLI

chasdr:
## Summary

Model selection in the Agent Assist skill kept its own copy of the LLM Gateway catalog fetch, so it drifted from the DataRobot CLI on pagination, active-model filtering, and which fields it surfaced. The CLI now exposes the same data through `dr llm-gateway list --output-format json`, and the skill's Pre-requisite Check already installs, upgrades, and authenticates `dr`. This switches the skill to that single source, which also picks up CLI-side catalog fixes for free.

## How it works

`list_llm_models.py` runs the CLI as a subprocess instead of calling `/genai/llmgw/catalog/` itself, and maps the CLI's envelope onto the `LLMModel` shape the skill already used, so `rehearsal.py` needs no change. The project's `.env` credentials are passed through so the listing stays pointed at the instance the project targets.

Two failure modes are handled deliberately. The CLI drops into an interactive login when it has no usable credentials, so the call closes stdin and carries a 90s timeout, turning a would-be hang into a clear error. And a `dr` that predates deployed-LLM support omits the per-entry `source` field, so a missing value is read as `gateway`, keeping the listing correct on an older CLI.

One thing to know when reviewing: the CLI honors passed credentials only once they verify, and otherwise falls back to its own stored profile. That makes a stale project `.env` produce a listing from a different DataRobot instance. The stacked CFX-6980 PR makes that visible by forwarding the CLI's log lines.

## Technical Changes

- `skills/datarobot-agent-assist/scripts/list_llm_models.py`: replace the urllib catalog request with a `dr llm-gateway list --output-format json` subprocess call. `fetch_cli_llms` owns the transport and error funnel, `fetch_llm_models` keeps its signature and gateway-only contract.
- `skills/datarobot-agent-assist/references/helper-scripts.md`: document that model selection now hard-depends on the CLI, and what to check when the listing looks like the wrong instance.
- `CHANGELOG.md`: entry under `[Unreleased]`.

Verified against a `dr` built from the CLI's merged `main`: 65 gateway models with descriptions and context sizes, `rehearsal.py`'s catalog still builds, and the error paths for a missing `dr`, a non-zero exit, and a malformed envelope all report rather than traceback. `ruff`, `mypy`, and 168 integration tests pass.

---

> [!NOTE]
> **Low Risk**
> Scoped to a design helper script and documentation; public output shape is preserved for downstream scripts like `rehearsal.py`.
> 
> **Overview**
> **Agent Assist design-time model selection** no longer calls the LLM Gateway catalog HTTP API directly. `list_llm_models.py` now runs `dr llm-gateway list --output-format json`, maps the `llms` envelope back to the existing `LLMModel` shape (including `datarobot/` prefixes), and keeps **gateway-only** entries via the CLI `source` field (missing `source` still counts as gateway for older `dr` builds).
> 
> Project `.env` credentials are injected into the subprocess (`DATAROBOT_ENDPOINT`, `DATAROBOT_API_TOKEN`, non-interactive mode). The call uses closed stdin and a 90s timeout so a CLI login prompt cannot hang the skill; missing `dr`, non-zero exit, and bad JSON all surface as `RuntimeError` messages. Docs in `helper-scripts.md` and `CHANGELOG.md` note that the CLI is now a **hard dependency** for model selection and that stale `.env` can yield a catalog from the CLI’s stored profile instead of the intended instance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85aa1fc7216903c3a20ca498ae6a6fefdd4fb3c5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "datarobot-agent-skills",
       "source": "./",
       "description": "DataRobot skills for AI/ML workflows — model training, deployment, predictions, feature engineering, monitoring, explainability, data preparation, App Framework CI/CD, and external agent monitoring.",
-      "version": "1.3.10"
+      "version": "1.3.11"
     }
   ]
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "datarobot-agent-skills",
   "description": "DataRobot skills for AI/ML workflows — model training, deployment, predictions, feature engineering, monitoring, explainability, data preparation, App Framework CI/CD, and external agent monitoring.",
-  "version": "1.3.10"
+  "version": "1.3.11"
 }
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "datarobot-agent-skills",
   "description": "A collection of DataRobot agent skills for model training, deployment, predictions, monitoring, and more.",
-  "version": "1.3.10",
+  "version": "1.3.11",
   "author": {
     "name": "DataRobot",
     "email": "carson.gee@datarobot.com"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Each entry should be prefixed with the affected skill folder name (for example,
 
 ## [Unreleased]
 
+- `datarobot-agent-assist`: Source model selection from the DataRobot CLI. `list_llm_models.py` now delegates to `dr llm-gateway list --output-format json` instead of calling the LLM Gateway catalog endpoint directly, so the skill and the CLI agree on what is available and pick up CLI-side catalog fixes for free. Project `.env` credentials are passed through to the CLI.
 - `datarobot-agent-assist`: Merge pre-coding spec gate into `pre-coding-checklist.md` Bootstrap step 2 (removes duplicate "Before Coding Begins" section). Add missing-spec recovery, path validation, session flags, workspace ask-don't-guess for Code, and journey deduplication. Trim `SKILL.md`: move CLI setup, helper scripts, plugin tool mapping, dress rehearsal prompt, path resolution, and agent_spec schema into references. Flow polish: Code-no-spec merges with Path resolution step 1, schema read hook in Spec Display, `design_to_code` guard and Windows stop in pre-coding, welcome menu resets `design_messy_cwd`.
 
 ## [1.3.9] - 2026-07-22
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,11 @@ Each entry should be prefixed with the affected skill folder name (for example,
 
 ## [Unreleased]
 
-- `datarobot-agent-assist`: Source model selection from the DataRobot CLI. `list_llm_models.py` now delegates to `dr llm-gateway list --output-format json` instead of calling the LLM Gateway catalog endpoint directly, so the skill and the CLI agree on what is available and pick up CLI-side catalog fixes for free. Project `.env` credentials are passed through to the CLI.
 - `datarobot-agent-assist`: Merge pre-coding spec gate into `pre-coding-checklist.md` Bootstrap step 2 (removes duplicate "Before Coding Begins" section). Add missing-spec recovery, path validation, session flags, workspace ask-don't-guess for Code, and journey deduplication. Trim `SKILL.md`: move CLI setup, helper scripts, plugin tool mapping, dress rehearsal prompt, path resolution, and agent_spec schema into references. Flow polish: Code-no-spec merges with Path resolution step 1, schema read hook in Spec Display, `design_to_code` guard and Windows stop in pre-coding, welcome menu resets `design_messy_cwd`.
+
+## [1.3.11] - 2026-07-29
+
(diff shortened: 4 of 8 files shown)


chasdr:
Note on the red `Skill quality evaluation` check: it is inherited from main, not introduced here.

This replaces an earlier note of mine that attributed the failure to the judge not being handed the `references/` directory. That is a real gap, but it is not what fails the build.

Main is already failing the same check on the same skill. [Run 30385080688](https://github.com/datarobot-oss/datarobot-agent-skills/actions/runs/30385080688) on `b54fba97`, the commit this branch is cut from, fails `test_skill_quality[datarobot-agent-assist]`. It has been red since #71 merged. #67 failed it three times on 2026-07-28 before going green, which turns out to be the most useful clue here.

**The gate does not fail on a `NEEDS WORK` verdict.** `verdict_passes()` accepts GOOD, NEEDS WORK and UNKNOWN, and fails only on `INCOMPLETE`. Every failing run on main, this PR, and #67 emitted `**NEEDS WORK**`.

**The check is flaky, not stuck red.** #67 went green at `ed9820fc`, and its only delta from the last failing SHA is one frontmatter `description` line, which touches none of the gaps the judge listed. It changed the file's MD5, forced a re-judge, and the re-judge passed on an otherwise identical skill body. Across 7 runs on effectively the same content: 6 fail, 1 pass. So on this skill the outcome is decided by incidental judge formatting rather than skill quality, which means red does not indicate a regression and green does not indicate the skill is fine.

What fails the build is `parse_verdict()` in `dr_agents_tester/pytest_plugin.py`:

1. The prompt asks for `### Overall verdict`. For this skill the judge prefixes the report with an H1 title and promotes every section a level, emitting `## Overall Verdict`.
2. The regex hardcodes level 3 (`r"###\s+Overall\s+verdict.*?\n(.+?)(?:\n|$)"`), so it does not match.
3. With no match, the function uppercases the **entire report** and scans that instead of one verdict line.
4. The label split `re.split(r"\s*[—–-]\s*", full_line, maxsplit=1)` fires on a bare hyphen with no surrounding whitespace, so the label is cut at the hyphen in `datarobot-agent-assist` and matches no verdict token.
5. The fallback then scans the whole report for `INCOMPLETE` first, and the word is present in body prose: "resolution failure path is incomplete", "tell the user the skill installation is incomplete".

The only variable that flips the build is which heading level the judge picked. The same report with `### Overall verdict` parses as NEEDS WORK and passes. That is also the likeliest reason `ed9820fc` went green: pytest captures stdout for passing tests, so that report is not recoverable, but a pass means `parse_verdict` did not return INCOMPLETE, which happens when either the heading lands at level 3 or the prose happens to omit the word.

This is the second false positive in that function; tester PR #4 was "fix: parse_verdict false positive", which is a good argument for unit tests over a third regex tweak. Worth noting for whoever picks this up: a green run does not verify a fix here. The six failing reports need capturing as fixtures and replaying through the corrected parser, because a single green CI run is indistinguishable from a lucky re-roll.

Editing `SKILL.md` here only changed the skill's MD5, which invalidated its entry in the `tests/e2e/skill_hashes.json` cache and made CI re-judge it. That is why the check surfaces on this PR and was quiet on the ones in between.

The fix belongs in `datarobot/datarobot-agent-tester`. Worth knowing that `uv.lock` pins that dependency at `a3c601ca` rather than resolving `@main` at install time, so it needs a lock bump here once the parser fix lands. Not attempting either in this PR.

The `references/` gap is still worth fixing, separately. `Skills.test()` reads only `skill_path.read_text()`, so verdicts on the four skills that use `references/` cite the missing files as gaps. That shapes verdict content, not the pass/fail outcome.

On #74: it stacks on this branch, so `lint`, `Trivy`, and this check do not run there at all. All three trigger only on `pull_request` with `branches: [main]`, and #74 targets `chas/CFX-6981`. Its green status is the absence of those checks, not a pass. It picks them up once this PR merges and GitHub retargets it to main.

carsongee (approved):
Before you merge, please bump the version numbers. Every merge is a release, so every PR needs to bump it

chasdr:
@carsongee on the version bump: #74 is stacked on this branch, so it retargets to main as soon as this merges. Since every merge is a release, keeping them separate means two releases and two bumps in a row.

Would you rather I fold #74 into this PR and bump 1.3.10 once, or keep them separate and bump each?

victorborshak (changes requested):
@chasdr Please make sure that CLI with respected changes is actually released. Otherwise the skill may be broken. 

Putting `Request changes` just to make sure you have seen this.

chasdr:
Superseded by #83, which is the combined PR off current `main` we discussed above.

@carsongee your version-bump ask is done there (1.4.3 to 1.5.0, minor per CONTRIBUTING since an existing skill gains a capability). Folding the two together means one release instead of two back-to-back, which was the question.

@victorborshak the release gate is clear: `dr` v0.2.79 (stable, 2026-08-03) carries both CFX-6985 and CFX-6984, and #83 also handles the case where someone is on an older `dr` that the template still accepts, since `dr llm-gateway list` only reports deployed LLMs from v0.2.79.

Leaving this open until #83 lands.

chasdr:
closing this. CFX-6981 turned out to be delivered by #79, which switched the listing to `dr llm-gateway list` on 2026-08-03. nothing left here to merge.

the remaining work is CFX-6980's `.env` half, which is #83.

PR #74, opened 2026-07-29
[CFX-6980] Support DataRobot-deployed LLMs in the Agent Assist skill

chasdr:
Stacked on #73. Review that one first. This PR's diff against `main` includes it.

## Summary

Agent Assist could only offer LLM Gateway models, so an environment where the gateway is disabled or empty had nothing to select. That is the normal shape of an on-prem install. DataRobot text-generation deployments are the answer there, and the CLI now lists them alongside gateway models. This makes them selectable end to end: the skill lists both sources, recommends across them, records the choice in `agent_spec.md`, and writes an `.env` that routes the template to the deployment instead of the gateway.

## How it works

Every deployed entry shares one model string, `datarobot/datarobot-deployed-llm`, and the deployment id carries the actual routing. So a deployed selection is a pair: `model` plus `llm_deployment_id`. The listing surfaces `source`, a human `label`, and `deployment_id` for exactly that reason, and `agent_spec.md` gains the optional second field. Passing the id to `setup_template.py` writes `LLM_DEPLOYMENT_ID`, `INFRA_ENABLE_LLM=deployed_llm.py`, and `USE_DATAROBOT_LLM_GATEWAY=0` alongside `LLM_DEFAULT_MODEL`, which is the contract the template's `infra/configurations/llm/deployed_llm.py` reads.

The half-configured case is the one worth guarding. `model` is always written to the spec while `llm_deployment_id` is the extra field that is easy to drop, and nothing downstream caught it: `dr dotenv setup` fills in the gateway default and keeps the deployed model name, so the run died much later at `pulumi up` with `Model 'datarobot-deployed-llm' not found in catalog`, which points nowhere near the missing id. `setup_template.py` now refuses that combination up front.

The dress rehearsal stays gateway-only, because it posts to the gateway's chat-completions endpoint and cannot address a deployment. `fetch_llm_models` keeps its gateway filter for that caller. On an instance with gateway models it substitutes one and says so. On a deployed-only instance the rehearsal cannot run at all, and now reports why instead of raising a traceback.

## Technical Changes

- `scripts/list_llm_models.py`: `fetch_llm_choices` returns both sources with `source`, `label`, and `deployment_id`. Table gains Source and Model/Deployment ID columns. Free-text deployment labels and descriptions are collapsed to one pipe-free line, since a newline in user-authored metadata split a table row apart. Off-contract CLI output (a null entry, a string `context_size`) now funnels through the module's `RuntimeError` channel instead of surfacing as a traceback. Forwards the CLI's log lines and echoes the requested endpoint, so a fallback to a different instance is visible.
- `scripts/setup_template.py`: new `--llm-deployment-id`, the deployed-model `.env` keys, and guards for both directions of the model/id mismatch.
- `scripts/rehearsal.py`: report a gateway-reachability failure instead of a traceback.
- `SKILL.md`, `references/agent-spec-schema.md`, `references/pre-coding-checklist.md`, `references/helper-scripts.md`: the `llm_deployment_id` field, recommending a deployed model when no gateway model exists, and passing the id through pre-coding step 9.

## Verification

Run against a `dr` built from the CLI's merged `main`, on a live instance with 65 gateway models and 9 deployments:

- Both sources list and render, with the deployed rows identified by label and deployment id.
- The deployed `.env` contract was checked against the template's `deployed_llm.py`, not from memory, and `dr dotenv setup --yes` was confirmed to preserve all three added keys rather than re-defaulting them.
- The guard fires only on the deployed model name without an id. Gateway models and a complete deployed pair both pass.
- A wrong-instance fallback is now legible: requesting a bogus endpoint shows the requested host and the host actually queried.
- `ruff`, `mypy`, and 168 integration tests pass. `test_gemini_extension_validate` fails locally for want of a `gemini` binary and fails the same way on `main`.

Known cost, not addressed here: the CLI call runs about 4.4s to 5.4s against roughly 1.9s to 2.9s for the old single catalog GET, since it verifies the token and pages both the catalog and deployments. `rehearsal.py --init` pays it too and then discards the deployed rows. A `--source` filter on the CLI side would fix that and belongs in its own ticket.

---

> [!NOTE]
> **Medium Risk**
> Changes the agent template setup path and `.env` LLM routing contract; misconfiguration is mostly guarded, but wrong instance/model selection could still mislead users before deploy.
> 
> **Overview**
> **Agent Assist** can now pick **DataRobot text-generation deployments** when the LLM Gateway is empty or disabled (typical on-prem), not only gateway catalog models.
> 
> `list_llm_models.py` returns both sources with `source`, `label`, and `deployment_id`; skill docs tell agents to prefer gateway models, fall back to deployed by `label`, and persist **`llm_deployment_id`** in `agent_spec.md` next to the shared sentinel `datarobot/datarobot-deployed-llm`. Pre-coding passes that id into **`setup_template.py --llm-deployment-id`**, which writes deployment routing keys (`LLM_DEPLOYMENT_ID`, `INFRA_ENABLE_LLM=deployed_llm.py`, `USE_DATAROBOT_LLM_GATEWAY=0`) and **fails fast** if the sentinel model is used without a deployment id.
> 
> Dress rehearsal stays gateway-only; **`fetch_llm_models`** errors clearly when only deployments exist, and **`rehearsal.py`** surfaces those failures without a traceback. CLI stderr (and the requested endpoint) is forwarded so a silent credential fallback to another instance is visible.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4de0f4d7b66aad3387e5c7f1aa193faa869df17. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chasdr:
Superseded by #83, which is the combined PR off current `main`.

This branch was cut from `b54fba9` before #67 moved the tree into `agent-assist-build/` and before #79 landed the CLI-sourced listing, the deployed entries, and the deployment chat route. #83 carries only the remaining delta rather than replaying this diff: the `.env` routing, `llm_deployment_id` in the spec, the docs, and the deployed-model fixes found while verifying.

Both bot findings here are handled in #83. The Copilot suggestion to fail fast when a deployment id is paired with a non-placeholder model is declined with a reason: `verify_llm(deployment_id=...)` in the template's `libllm.py` returns before any catalog check and treats the model string as a label the endpoint ignores, so failing would break a supported setup.

Leaving this open until #83 lands.

chasdr:
closing this. it was cut before #67 moved the skill into `agent-assist-build/` and before #79 landed the listing and the rehearsal routing, so rebasing it would mostly be reverting other people's work.

#83 carries what's actually left off current main.

PR #75, opened 2026-07-29
[DO NOT MERGE] ci: probe stacked-PR check coverage

chasdr:
**Throwaway. Do not merge, do not review.** This branch and its stacked child exist to measure CI behavior and will be closed and deleted once the three observations below are recorded.

Diagnostic for two problems on main, neither introduced by any open PR:

1. `Skill quality evaluation` is red on main since #71 merged, and on #67 and #73. Cause is a parser defect in `parse_verdict()` in `datarobot-agent-tester`, not the skill content. Detail in [this comment on #73](https://github.com/datarobot-oss/datarobot-agent-skills/pull/73#issuecomment-5119476553).
2. `lint`, `Trivy`, and `Skill quality evaluation` all trigger on `pull_request` with `branches: [main]`, so a stacked PR gets no checks at all and its green status is the absence of checks rather than a pass.

## What this branch does

- Drops the `branches: [main]` filter from the `pull_request` trigger in `lint.yml`, `test-skills-e2e.yml`, and `trivy-scan.yml`. The `push: branches: [main]` filters are untouched.
- One-word copy fix in `skills/datarobot-agent-assist/SKILL.md`, needed only to change the file's MD5 so the judge's hash cache re-evaluates it and to satisfy the e2e `paths:` filter.

## What is being measured

- Whether the stacked child PR gets `lint`, `Trivy`, and `Skill quality evaluation`. `pull_request` resolves the workflow from the merge ref, so the child is evaluated under the widened filter on this branch.
- Whether `Skill quality evaluation` is a required check. Rulesets, classic branch protection, and the GraphQL `branchProtectionRules` all read empty at my permission level, yet #73 reports `mergeStateStatus: BLOCKED`, so something is configured that I cannot see.
- Which heading level the judge emits for its verdict section this run. Four prior runs all emitted `## Overall Verdict` where the prompt asked for `### Overall verdict`, which is the specific mismatch that fails the build.

If the filter change measures out, it gets re-opened as its own reviewable PR off main rather than merged from here.

PR #76, opened 2026-07-29
[DO NOT MERGE] ci: probe child, stacked on the filter change

chasdr:
**Throwaway. Do not merge, do not review.** Stacked on #75, closed and deleted with it.

This PR is the measurement. It targets `chas/ci-probe-parent`, not `main`. Under the filters currently on main it would receive no checks at all, the same as #74. #75 drops the `pull_request` branch filter, and because `pull_request` resolves the workflow from the merge ref, this PR is evaluated under that widened filter.

So the check list here answers the question directly: if `lint`, `Trivy`, and `Skill quality evaluation` appear below, widening the filter gives stacked PRs real coverage.

Content is a one-word copy fix in `skills/datarobot-discover/SKILL.md`, present only to satisfy the e2e `paths:` filter and to force a re-judge of a skill that does not use `references/`.

### DIFF
--- a/skills/datarobot-discover/SKILL.md
+++ b/skills/datarobot-discover/SKILL.md
@@ -10,7 +10,7 @@ description: >-
 # Discover DataRobot Resources
 
 Fetch DataRobot catalog and present matching resources for the user's
-task. Covers resource types including Global MCP Server. Bring resources once DataRobot publishes them without skill update.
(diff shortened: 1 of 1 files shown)


chasdr:
Measurement recorded. Stacked PRs do get full coverage once the pull_request branch filter is dropped: this PR ran lint, Trivy, Trivy Secret Scanner, and Skill quality evaluation against a non-main base, where it would otherwise have had route alone. Closing; the filter change will come back as its own PR off main.

PR #88, opened 2026-08-19
Stop helper scripts inventing a DataRobot endpoint they cannot authenticate

mhauskn-dr:

## Summary

Helper scripts and SKILL.md examples build their client like this:

```python
dr.Client(
    token=os.getenv("DATAROBOT_API_TOKEN"),
    endpoint=os.getenv("DATAROBOT_ENDPOINT", "https://app.datarobot.com"),
)
```

There is no fallback for the token, so when `DATAROBOT_*` is not in the environment this passes an endpoint alongside a `None` token. The SDK requires both to come from the same source and refuses before making a request:

```
ValueError: Token must be specified if endpoint is specified
```

Removing the default lets `dr.Client()` run its own resolution order. One line, 14 files.

## Rationale

**This is the state `datarobot-setup` produces.** `dr auth login` writes `~/.config/datarobot/drconfig.yaml` and sets no environment variables. So a user who follows our own setup skill, with entirely valid credentials, runs a helper script and is told their token is missing. The natural next step — reissuing the API token — cannot help, because the token was never the problem.

**The default value is also wrong on its own terms.** The API lives at `https://app.datarobot.com/api/v2`. Supplying the un-suffixed URL *with* a valid token fails differently:

```
ClientError: Server did not reply with an API version
```

So the default could not work in either direction: without a token it raises, with a token it points at the web app instead of the API.

**The SDK already does this correctly.** `dr.Client()` resolves explicit arguments, then environment, then `drconfig.yaml`, and defaults to the public cloud endpoint with the right suffix. The line duplicated a decision the SDK owns, one layer too high, and got it wrong. Passing `None` is how you tell the SDK to decide.

Verified against `datarobot` 3.18.0:

| Call | Result |
| --- | --- |
| `dr.Client(token=None, endpoint=None)` | OK — resolves to `https://app.datarobot.com/api/v2` |
| `dr.Client(token=None, endpoint="https://app.datarobot.com")` | `ValueError: Token must be specified if endpoint is specified` |
| `dr.Client(token=<valid>, endpoint="https://app.datarobot.com")` | `ClientError: Server did not reply with an API version` |

## No configuration regresses

| User's setup | Before | After |
| --- | --- | --- |
| `DATAROBOT_ENDPOINT` set | works | unchanged — `os.getenv` still returns it |
| `drconfig.yaml` only | `ValueError` | works |
| neither | confusing `ValueError` | clearer missing-credentials error |

## Scope

**This removes the first blocker; it does not make the scripts work end to end.** Several still fail afterwards on separate SDK mismatches — methods that do not exist, calls with the wrong arguments. Those are being handled separately, and are deliberately not mixed in here. This change is worth landing on its own because it is the one failure every script shares, it happens before anything else can run, and it misdirects whoever hits it.

---

> [!NOTE]
> **Low Risk**
> Documentation and example-script credential wiring only; no runtime service or auth logic changes beyond aligning with SDK defaults.
> 
> **Overview**
> **Release 1.5.2** bumps shared plugin metadata (`package.json`, Claude/Cursor/Gemini manifests) and documents the fix in `CHANGELOG.md`.
> 
> Across **data preparation, feature engineering, model training/deployment/monitoring, and predictions** skills, `dr.Client()` initialization no longer passes `endpoint=os.getenv("DATAROBOT_ENDPOINT", "https://app.datarobot.com")`. Examples in `SKILL.md` and the same pattern in helper scripts (`upload_dataset.py`, training scripts, prediction scripts) now use **`endpoint=os.getenv("DATAROBOT_ENDPOINT")` only**.
> 
> That change fixes a common failure when users authenticate via **`dr auth login`** / `drconfig.yaml` without env vars: supplying a default endpoint with a `None` token triggers **`ValueError: Token must be specified if endpoint is specified`**, and the old default URL was not the API base path anyway. With no explicit endpoint, **`dr.Client()`** applies its normal resolution (args → env → config file → public cloud API).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ab0b1f5a17176f497cd4caedd97b9b5b018a99e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skills/datarobot-predictions/SKILL.md
+++ b/skills/datarobot-predictions/SKILL.md
@@ -352,7 +352,7 @@ import os
 # Initialize client with API credentials
 client = dr.Client(
     token=os.getenv("DATAROBOT_API_TOKEN"),
-    endpoint=os.getenv("DATAROBOT_ENDPOINT", "https://app.datarobot.com"),
+    endpoint=os.getenv("DATAROBOT_ENDPOINT"),
 )
 ```
 
--- a/skills/datarobot-predictions/SKILL.md
+++ b/skills/datarobot-predictions/SKILL.md
@@ -148,7 +148,7 @@ import datarobot as dr
 import pandas as pd
 from datarobot_predict.deployment import predict as dr_predict
 
-dr.Client(token=..., endpoint=...)
+dr.Client()
 deployment = dr.Deployment.get("abc123")
 
 result = dr_predict(
@@ -244,15 +244,11 @@ Claude can run these scripts directly or use them as reference when writing code
 ### Pattern 1: Get deployment features and make single prediction (optionally with explanations)
 ```python
 import datarobot as dr
-import os
 import pandas as pd
 from datarobot_predict.deployment import predict as dr_predict
 
 # Initialize client
-dr.Client(
-    token=os.getenv("DATAROBOT_API_TOKEN"),
-    endpoint=os.getenv("DATAROBOT_ENDPOINT"),
-)
+dr.Client()
 
 deployment = dr.Deployment.get("abc123")
 
@@ -276,12 +272,9 @@ print(result.dataframe.to_dict(orient="records"))
 ```python
 import datarobot as dr
 import pandas as pd
-import os
 
 # Initialize client
-client = dr.Client(
-    token=os.getenv("DATAROBOT_API_TOKEN"), endpoint=os.getenv("DATAROBOT_ENDPOINT")
-)
+dr.Client()
 
 # Get deployment features
 deployment = dr.Deployment.get("abc123")
@@ -347,13 +340,9 @@ pip install datarobot
 
 ```python
 import datarobot as dr
-import os
 
 # Initialize client with API credentials
-client = dr.Client(
-    token=os.getenv("DATAROBOT_API_TOKEN"),
-    endpoint=os.getenv("DATAROBOT_ENDPOINT"),
-)
+dr.Client()
 ```
 
 ### Environment Variables
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "datarobot-agent-skills",
       "source": "./",
       "description": "DataRobot skills for AI/ML workflows — model training, deployment, predictions, feature engineering, monitoring, explainability, data preparation, App Framework CI/CD, and external agent monitoring.",
-      "version": "1.5.1"
+      "version": "1.5.2"
     }
   ]
 }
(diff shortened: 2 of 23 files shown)


skills/datarobot-predictions/SKILL.md
```diff
@@ -352,7 +352,7 @@ import os
 # Initialize client with API credentials
 client = dr.Client(
     token=os.getenv("DATAROBOT_API_TOKEN"),
-    endpoint=os.getenv("DATAROBOT_ENDPOINT", "https://app.datarobot.com"),
+    endpoint=os.getenv("DATAROBOT_ENDPOINT"),
```
carsongee: I think these should all just be `dr.Client()`. It will pull everything in properly to configure things

carsongee (commented):
Great call out, and fun proof that the evaluation framework is working. I think the fix is a bit wrong since `dr.Client()` with no arguments will figure this all out automatically. Please use the "Ready for Review" label when you've addressed the feedback, and we're ready to merge!

skills/datarobot-predictions/SKILL.md
mhauskn-dr: Good call — done in cfbd8e9. This is better than what I had, in a way I hadn't appreciated: it fixes the class rather than the instance.

Verified against `datarobot` 3.18.0 that `dr.Client()` and the explicit form resolve identically with both variables set, with only one set, and with neither. In the one case they differ — `DATAROBOT_ENDPOINT` set but no token — bare gives the better message ("Endpoint and token must come from the same source" rather than "Token must be specified if endpoint is specified").

It also caught a site my original sweep missed. `datarobot-model-explainability` used a different spelling:

```python
token=os.environ["DATAROBOT_API_TOKEN"],
endpoint=os.environ.get("DATAROBOT_ENDPOINT", "https://app.datarobot.com/api/v2"),
```

That default is correctly suffixed, so it wasn't the bug I was fixing — but it still fails on a drconfig-only machine, just with `KeyError` instead of `ValueError`. Fixing by pattern-match has to enumerate spellings and will keep missing one; `dr.Client()` can't be got wrong.

I also included **`AGENTS.md`**, since it's where the pattern comes from — it demonstrates the redundant form to every agent that reads it, and `CLAUDE.md` symlinks to it, so the leaf sites would grow back without it. Shout if you'd rather that were separate.

Two things fell out: the assigned `client` variable was never referenced at any of these sites, and `import os` became unused wherever those lookups were its only use. Net −118/+36.

Left alone: the two `create_use_case.py` / `create_shell_deployment.py` calls in `datarobot-external-agent-monitoring`, which take `token` and `endpoint` as function parameters — a real use of the arguments.

mhauskn-dr:
@carsongee - I can't edit the labels on this PR - so I'm not sure there's any way to apply the "Ready for Review" label? I was able to hit the button to request your re-review. Thanks for the feedback!

carsongee (approved):

PR #89, opened 2026-08-24
docs(predictions): install datarobot-predict — examples import it but setup omits it

miyamamoto:
## Problem

The `datarobot-predictions` skill's real-time prediction and prediction-explanation examples import `datarobot_predict`:

```python
from datarobot_predict.deployment import predict as dr_predict   # SKILL.md
```

…but the setup instructions install only `datarobot`:

```bash
pip install datarobot
```

`datarobot-predict` is a **separate PyPI package** and is **not** a dependency of `datarobot`, so following the skill exactly as written fails at import:

```
ModuleNotFoundError: No module named 'datarobot_predict'
```

This is environment-independent (pure Python import) — it happens on every platform, Cloud or self-managed.

## Reproduce (clean venv)

```bash
python -m venv env && ./env/bin/pip install datarobot        # exactly what the skill says
./env/bin/python -c "from datarobot_predict.deployment import predict"
#  -> ModuleNotFoundError: No module named 'datarobot_predict'
./env/bin/pip install datarobot-predict
./env/bin/python -c "from datarobot_predict.deployment import predict"   # OK
```

## Fix

Add `datarobot-predict` to the two `pip install` lines in the skill's setup so the documented examples run as written. Two-line docs change, no behavior change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Docs-only install-line change; no runtime, auth, or data-handling code is modified.
> 
> **Overview**
> Fixes skill setup so documented prediction examples can actually import `datarobot_predict`.
> 
> Both `pip install` snippets in `SKILL.md` now include `datarobot-predict` as well as `datarobot`, matching the real-time/explanation samples that `from datarobot_predict.deployment import predict`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e67d00c89ca6f05d2389a01f28bfa0f7e24cb18. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
       "name": "datarobot-agent-skills",
       "source": "./",
       "description": "DataRobot skills for AI/ML workflows — model training, deployment, predictions, feature engineering, monitoring, explainability, data preparation, App Framework CI/CD, and external agent monitoring.",
-      "version": "1.5.1"
+      "version": "1.5.2"
     }
   ]
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,5 @@
 {
   "name": "datarobot-agent-skills",
   "description": "DataRobot skills for AI/ML workflows — model training, deployment, predictions, feature engineering, monitoring, explainability, data preparation, App Framework CI/CD, and external agent monitoring.",
-  "version": "1.5.1"
+  "version": "1.5.2"
 }
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "datarobot-agent-skills",
   "description": "A collection of DataRobot agent skills for model training, deployment, predictions, monitoring, and more.",
-  "version": "1.5.1",
+  "version": "1.5.2",
   "author": {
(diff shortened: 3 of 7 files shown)


carsongee:
Please update the pins and add to the changelog! @danna-naser can you or a team member of yours take a quick double check that we wouldn't want to fix this by changing skill code to not need the library at all? Thanks!

miyamamoto:
Thanks @carsongee! Done — bumped the shared plugin version to 1.5.2 across `package.json`, `.claude-plugin/{plugin,marketplace}.json`, `.cursor-plugin/plugin.json` and `gemini-extension.json`, and added the changelog entry.

On dropping the library entirely: totally open to that if the team prefers. For context on the tradeoff (@danna-naser) — batch scoring works with the plain `datarobot` SDK (`deployment.predict_batch(...)` / `BatchPredictionJob`), but the skill's single-row **real-time** path and the per-row prediction **explanations** currently rely on `datarobot_predict.deployment.predict`. If you'd rather remove the dependency, I can rework the examples onto the plain SDK where it's supported and flag anything that would lose real-time / explanation coverage.

We're all on the same team here, so please don't hesitate to reach out.

carsongee (approved):
Thanks!

PR #96, opened 2026-08-27
Audit and repair skill issues

mhauskn-dr:

## Summary

Repo-wide audit and repair of the skill library against the DataRobot Python SDK **3.18.0** (current release) and the live API, followed by four rounds of behavioral testing with recorded agent runs. **116 verified defects fixed across 12 of the 13 skills** (only `datarobot-setup` was clean): SDK signature mismatches, nonexistent attributes and methods, deprecated APIs, semantic bugs that returned wrong answers without raising, missing dependencies, and doc gaps that recorded agent runs demonstrably tripped over.

- 4 commits: `f8e661a` (audit fixes), `bb02176` / `11d3c29` / `aa7c685` (behavioral round 2/3/4 follow-ups)
- 42 files, +537/−283; plugin version bumped to **1.5.6** with a full CHANGELOG entry
- Behavioral outcomes across the four rounds (6 scenarios): friction signals **16 → 1**, skill-logic failures **13 → 0**, delivered-script SDK defects **1 → 0**, wall time **1897s → 1160s (−39%)**, turns **103 → 75**, pass rate 6/6 in every verified round
- `task lint` green (296 tests) at every commit; no server-side resources created or modified at any point — all verification was introspection plus read-only API calls

## Rationale

These skills exist so coding agents can one/few-shot DataRobot tasks. When a skill documents a call that does not exist in the SDK, every agent that loads it *starts* by executing a failure and then burns turns diagnosing a problem the skill created — and the worst defects never raise at all: they return confidently wrong answers.

Recorded behavioral runs (13 baseline runs across 6 representative scenarios, catalogued in `datarobot-agent-tester/docs/behavioral-run-errors.md`) made the cost concrete:

- **66% of all observed errors traced to skill documentation**, not the agent or the SDK. Four defects were copy-pasted across multiple skills and hit every run that touched them.
- The costliest single error — a 422 for a missing `defaultPredictionServerId` that the SDK types `Optional[str]` and no local introspection can reveal — fired only *after* 11 minutes of paid Autopilot, twice.
- The silent failures were worse than the crashes: a model-ranking snippet compared per-partition **dicts** and silently returned an arbitrary "best" model on regression projects; a monitoring report answered **0 predictions** for a deployment that had served 60 inside the reported window; a feature-impact report claimed a feature "accounts for ~65% of the predictive signal" — a number that measures nothing.

Static conformance checking (PR #86) catches roughly half of this class — signature arity and names — and none of the attribute reads, deprecations, return-shape errors, or semantics. That is why this PR pairs a full SDK audit with live verification and behavioral validation.

## Methodology

Every fix in this PR passed through a five-stage pipeline in which **no claim was written into a skill without machine evidence** — `inspect.signature`/`hasattr`/SDK source for static facts, read-only live API calls for server-side behavior.

1. **Audit** — one auditor per skill resolved every SDK reference (calls with kwargs, attribute reads, module members, imports, documented return shapes, install instructions) against a clean install of `datarobot` 3.18.0, plus live read-only GETs against `app.datarobot.com` for behavior that introspection cannot see (e.g., which keys `deployment.model` actually carries per deployment type). Two pre-existing defect catalogues were treated as leads and independently re-verified rather than trusted — several of their claims were refuted in the process. Result: 101 candidate findings.
2. **Adversarial verification** — an independent skeptic per skill attempted to *refute* each finding, and validated each proposed fix by introspecting the replacement API before any edit. Result: **100 confirmed, 2 refuted, 2 downgraded**, and 4 fixes amended — including one whose original version would itself have introduced a new error (a `datarobot_predict` serverless caveat that is false for ≥1.13.5).
3. **Fix pass** — per-skill edits bound to a shared canonical-patterns contract so cross-skill idioms stay consistent (two-arg `Model.get`, `deployment.get_features()` dicts, registered-model deployment flow, upload-then-`modify(name=…)`). Mechanical gates: `py_compile` + ruff + mypy on every script, `ast.parse` on every edited python fence, `yaml.safe_load` on YAML, and live read-only execution of the fixed helper scripts against fixture deployments (both native and custom-model).
4. **Post-fix review** — three independent reviewers over the final diff: every *newly written* SDK claim re-verified, every edited file re-read whole for internal consistency, and a traceability ledger confirming each confirmed finding maps to an edit and each refuted finding was left untouched.
5. **Behavioral validation loop** — four rounds of scenario runs against the evolving branch (baseline → `f8e661a` → `bb02176` at k=2 → `11d3c29` at k=1 + a k=4 focus run). Each round's findings were themselves re-verified against the live API before being acted on (one round-report claim was corrected in the process: the raw `targetDrift` default window is last-7-days, not all-time). The rounds also produced transferable authoring rules now applied across the library: output constraints beat prohibitions (agents abstain correctly when the *output* rule names the constraint), example-code constants compete with prose tables for authority, and shell operators in snippets don't survive agent rewrites while prose decisions do.

## Appendix — every defect and its fix

Commit map: `f8e661a` = SDK audit fixes · `bb02176` = behavioral round 2 · `11d3c29` = round 3 · `aa7c685` = round 4. Paths in the **Where** column are relative to `skills/<skill>/`. Behavioral evidence: `docs/behavioral-run-2-results.md` … `run-4` (round reports), `datarobot-agent-tester/docs/behavioral-run-errors.md` (baseline catalogue).

Behavioral trajectory over the four rounds:

| | baseline | round 2 `f8e661a` | round 3 `bb02176` | round 4 `11d3c29` |
|---|---|---|---|---|
| pass | 6/6 | 6/6 | 11/12 | **6/6** |
| friction signals | 16 | 3 | 4¹ | **1** |
| — skill-logic | 13 | 1 | 0 | **0** |
| delivered-script SDK defects | n/a | 1 | 0 | **0** |
| wall (s) | 1897 | 1251 | 1251 | **1160** |

¹ round 3 ran on a freshly rebuilt venv and a stricter detector, exposing install friction the earlier rounds structurally could not see.

<details>
<summary><b>datarobot-predictions</b> — 19 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Taught `dr.Model.get(deployment.model["id"])` — the real signature is `(project, model_id)`, so every feature-lookup path crashed with `TypeError` before doing anything | Dropped the `Model` lookup entirely; feature metadata now comes from `deployment.get_features()` (also works for custom-model deployments, whose `deployment.model` lacks `project_id`) | `SKILL.md`, `scripts/generate_prediction_data_template.py`, `scripts/get_deployment_features.py`, `scripts/validate_prediction_data.py` — `f8e661a` |
| Taught the invented `model.get_features()` and then read `f.name`/`f.feature_type` object attributes — `Model` has no such method, and the real call returns plain dicts | Uses `deployment.get_features()` with dict access (`f["name"]`/`f["feature_type"]`/`f["importance"]`); the now-redundant `model.get_feature_impact()` importance lookups were removed since importance is already in the dicts | `SKILL.md`, `scripts/generate_prediction_data_template.py`, `scripts/get_deployment_features.py`, `scripts/validate_prediction_data.py` — `f8e661a` |
| Read `model.target_name` — no such attribute on `Model` (absent even on a live-fetched instance), so filtering the target out of the feature list crashed with `AttributeError` | Reads `deployment.model.get("target_name")` — present on both native and custom-model deployments, `.get()` because it is `None` for unsupervised/anomaly deployments | `SKILL.md`, `scripts/generate_prediction_data_template.py`, `scripts/get_deployment_features.py`, `scripts/validate_prediction_data.py` — `f8e661a` |
| Read `model.target_type` — no such attribute on `Model` | Reads `deployment.model["target_type"]` (verified present on all sampled deployments) | `scripts/get_deployment_features.py` — `f8e661a` |
| Read `model.project_name` in the template's metadata comment — no such attribute, and the key is also absent from `deployment.model` on custom-model deployments | Reads `deployment.model.get("project_name", "unknown")` | `scripts/generate_prediction_data_template.py` — `f8e661a` |
| Taught `dr.BatchPredictionJob.score(deployment_id=...)` — no such keyword exists (`TypeError`); the first parameter is `deployment`, accepting a `Deployment` or id string | Calls `score(deployment=deployment, ...)`; the `localFile` intake/output settings dicts were verified valid as written and kept | `SKILL.md` — `f8e661a` |
| Taught classmethod-style `dr.BatchPredictionJob.download(job.id)` — `download` is an instance method requiring a file object, so the idiom crashes whenever actually invoked | Line removed: with a `localFile` output path, `score()` has already written the results file before returning; `dr.BatchPredictionJob.score_to_file(...)` is shown as the equivalent one-liner | `SKILL.md` — `f8e661a` |
| Compared batch-job status against lowercase `"completed"` — server statuses are uppercase (`COMPLETED`), so the guard silently never fired and agents wasted turns polling a job whose output file already existed | Deleted the whole monitor/download block and documented that `score()` with a `localFile` output path blocks until scoring finishes and the file is written | `SKILL.md` — `f8e661a` |
| Called the invented `project.get_time_series_info()` — an `AttributeError` the surrounding `except dr.errors.ClientError` did not catch, crashing the script for every time-series deployment | Uses `dr.DatetimePartitioning.get(project_id)` for the datetime column, forecast window, and series-id columns, reached via `deployment.model.get("project_id")` with a guard for custom-model deployments that lack it | `scripts/get_deployment_features.py` — `f8e661a` |
| Documented `deployment.get_features()` without saying it returns plain dicts or what `importance` means — the skill's own Pattern 2 treated the return as attribute-bearing objects, and behavioral runs mis-ranked template columns by the field | Documents the exact dict keys (`name`, `feature_type`, `importance`, `date_format`, `known_in_advance`), dict access, and that `importance` is a model-independent feature/target association score that can be negative and is NOT Feature Impact | `SKILL.md` — `f8e661a` |
| Claimed `max_explanations=0` "disables explanations" unconditionally — with `explanation_algorithm="shap"`, `0` computes ALL explanations, so the documented default silently returned every explanation column | Documents the split semantics: `0` disables only for `xemp`; with SHAP, `0` (or `"all"`, SHAP-only) computes all explanations | `SKILL.md`, `scripts/make_prediction.py` — `f8e661a` |
| `make_prediction.py`'s docstring promised a nested `explanations` list of `{feature, value, strength, qualitative_strength}` per row — the script returns raw dataframe records, so parsing `row["explanations"]` got `KeyError` | Docstring now describes the real flat `EXPLANATION_<n>_FEATURE_NAME`/`_ACTUAL_VALUE`/`_STRENGTH`/`_QUALITATIVE_STRENGTH` columns | `scripts/make_prediction.py` — `f8e661a` |
| Documented the `DATAROBOT_ENDPOINT` default as `https://app.datarobot.com` — missing the required `/api/v2` path, so exporting the documented value fails every API call | Documents that the endpoint must include the `/api/v2` path (e.g. `https://app.datarobot.com/api/v2`) | `SKILL.md` — `f8e661a` |
| Template script filled every categorical with a `"sample_category"` placeholder and the error-handling advice said "use training data sample" without naming an API, leaving agents to invent invalid category values | Fills categoricals with real training levels via `dr.Feature.get(project_id, name).get_histogram()` (cycled across rows, listed in a "Valid values" header comment; placeholder fallback for custom-model deployments), and the error-handling section now names that API | `SKILL.md`, `scripts/generate_prediction_data_template.py` — `bb02176` |
| Showed the `score_to_file` one-liner with bare positional arguments, forcing agents to guess the parameter meanings (round 2's only golden-run failure) | Example shows the explicit `intake_path=`/`output_path=` keywords | `SKILL.md` — `bb02176` |
| Both install sites taught bare `pip install datarobot datarobot-predict`, which failed in behavioral runs with `command not found: pip` and PEP 668 refusals | Uses interpreter-anchored `python -m pip install ...` with a `uv pip install` fallback note for pip-less venvs and PEP 668 systems | `SKILL.md` — `bb02176` |
| Presented `uv pip` as a separate fallback step after `python -m pip`, costing two wasted tool calls per install in pip-less environments (round 3) | Single chained one-liner `python -m pip install ... \|\| uv pip install ...` resolves the right installer in one call | `SKILL.md` — `11d3c29` |
| Template script filled numeric columns with `0.0` placeholders, producing out-of-distribution template rows (round 3) | Fills numerics with the feature's training median and lists min/max ranges in the header comments, via the same `dr.Feature.get` stats already fetched for categorical levels (placeholder fallback for custom-model deployments) | `scripts/generate_prediction_data_template.py` — `11d3c29` |
| The installer decision lived only in shell syntax — round 4 showed agents that rewrite the install command (extra packages, `-r requirements.txt`) keep the ordering but drop the `\|\|` operator | Prose next to both install commands states the rule: if `python -c "import pip"` fails, go straight to `uv pip install` | `SKILL.md` — `aa7c685` |

**Considered and deliberately not changed:**

- `SKILL.md:144` (add a note that `datarobot_predict` "requires `deployment.default_prediction_server`") — REFUTED: `datarobot_predict` >= 1.13.5 has an explicit serverless branch before that `ValueError` (`_is_serverless_deployment` routes serverless deployments through the API endpoint with Bearer auth), so the suggested caveat would itself have been a new doc error; no edit made.
- sdk-discrepancies.md D2b ("`BatchPredictionJob` has no `.status`", citing `SKILL.md:315,318`) — refuted: instances do carry `.status` (set in `AbstractJob.__init__`; live value `'ABORTED'`), only class-level `hasattr` is False; the real defect at that site was the lowercase `"completed"` comparison, which is what got fixed.

</details>

<details>
<summary><b>datarobot-workload-api</b> — 16 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Taught debug-flow step 1 to read `statusDetails.logTail`/`statusDetails.conditions` from `GET /workloads/{id}/` — fields that never exist on the response (`WorkloadFormatted` is `additionalProperties: false`), so the script's step-1 log scan was a permanent silent no-op and `wait_for_running.py` dumped an always-null `statusDetails` on failure. | Step 1 reads `status` only; error lines come from `GET /otel/workload/{id}/logs/?level=error` (scanning `message`), conditions from the proton `statusDetails` endpoint; the script scans OTel error logs and `wait_for_running.py` points at the diagnose script instead. | `SKILL.md`, `scripts/diagnose_workload.py`, `scripts/wait_for_running.py` — `f8e661a` |
| Taught flagging workload events on `type: Warning` and `reason`/`message` fields that `WorkloadEvent` does not have (real keys: `eventType`, `details`), so the script's flagged-events check silently matched nothing and the report's promised "flagged events" section never rendered. | Filters on `eventType` containing error keywords (e.g. `"Replacement Errored"`) and reads the `details` object as evidence. | `SKILL.md`, `scripts/diagnose_workload.py` — `f8e661a` |
| Taught `DELETE /workloads/{id}/stats/?metricName=<name>` zeroes a single metric's history — no `metricName` param exists, so the server ignores it and the call silently wipes ALL stats in scope (a destructive divergence). | Documents `DELETE /workloads/{id}/stats/` as clearing the full stats history, scoped only by `protonId`/`startTime`/`endTime`, with no per-metric delete. | `SKILL.md` — `f8e661a` |
| Autoscaling examples and prose put `minCount`/`maxCount` inside the policy object — `AutoscalingPolicy` allows only `scalingMetric`/`target` (`additionalProperties: false`), so the example body failed request validation. | Uses `minReplicaCount`/`maxReplicaCount` on the `autoscaling` object (per `AutoscalingProperties`) with policies holding only `scalingMetric`/`target`, at all four sites. | `SKILL.md`, `references/lifecycle-flows.md`, `references/schema-reference.md` — `f8e661a` |
| Taught that OTel trace `duration` is nanoseconds on summaries AND spans (divide by 1e6 for ms) — live values show trace list/detail duration is SECONDS and span duration is MICROSECONDS, so displayed latencies were silently off by 1e9 and 1e3. | Documents units per level: trace list/detail duration in seconds (x1000 for ms), span duration in microseconds (/1000 for ms), span `startTime` as epoch ms. | `SKILL.md` — `f8e661a` |
| Metric conversion taught `nanocores`→cores via `/1_000_000` (1000x overstatement — 1 core = 1e9 nanocores) and treated `percentagedecimal` as already % (100x misreport), while ignoring the `mebibytes` and null units entirely. | Branches on each record's `unit` field: bytes /1024**2, mebibytes as-is, nanocores /1_000_000_000, percentagedecimal x100, percentage as-is, null shown raw. | `SKILL.md` — `f8e661a` |
| Manual code-sync taught PATCHing `codeRef` directly onto the container — `Container` is `additionalProperties: false` with no such property (it lives at `imageBuildConfig.codeRef`), so the patch was rejected or dropped and the build never found the source. | PATCHes the container's `imageBuildConfig.codeRef` with the catalog id + version id; the prose and state-map row name the nested location explicitly. | `references/code-to-workload.md` — `f8e661a` |
| Status docs listed nonexistent workload statuses `updating` (rolling redeploys) and `failed` (unrecoverable), plus proton-only `interrupted` — agents matching on them could never fire, and `wait_for_running.py` carried `failed` as dead terminal-match code. | Status tables pruned to the real `WorkloadStatus` enum; rolling redeploys observed via the `candidate` proton, terminal-failure matching uses `errored`/`terminated` only, `interrupted` noted as proton-only. | `SKILL.md`, `references/status-vocabulary.md`, `scripts/wait_for_running.py` — `f8e661a` |
| `wait_for_build.py` treated `FAILED` as the only terminal build failure — a `CANCELLED` build spun the full 1800s timeout and exited 3 (timeout) instead of 2 (failure), and every build-status sequence in the docs omitted `CANCELLED`. | `FAILURE = {"FAILED", "CANCELLED"}` in the poller; `CANCELLED` documented as terminal across all build-status sequences (and schema-reference's "BUILT or COMPLETED both terminal-success" line corrected to only-`COMPLETED`). | `scripts/wait_for_build.py`, `SKILL.md`, `references/status-vocabulary.md`, `references/code-to-workload.md`, `references/lifecycle-flows.md`, `references/schema-reference.md` — `f8e661a` |
| Scripts referenced nonexistent pre-consolidation sibling skills: `wait_for_replacement.py` printed a diagnose command at `../../datarobot-workload-debug/` (file-not-found at the failure-recovery moment) and `diagnose_workload.py` recommended loading `datarobot-workload-telemetry`/`-artifacts`/`-management`. | The diagnose command resolves `diagnose_workload.py` via `Path(__file__).parent` (works from any cwd); recommendations point at this skill's own SKILL.md sections (1: settings PATCH, 3: OTel logs, 4: artifact lifecycle). | `scripts/wait_for_replacement.py`, `scripts/diagnose_workload.py` — `f8e661a` |
| Named `${DATAROBOT_ENDPOINT}/openapi.yaml` as the schema source of truth — that spec contains zero workload/artifact paths or schemas (verified on an org with workloads enabled), so every documented `yq` lookup silently returned null. | Points spec lookups at the docs-site spec `https://docs.datarobot.com/en/docs/api/reference/public-api/openapi.yaml` (which has all 17+ workload paths and schemas), noting the endpoint-served spec lacks the namespace. | `SKILL.md`, `references/schema-reference.md` — `f8e661a` |
| Example `yq` lookup used path key `"/workloads/{workloadId}/"` — neither spec has that key (real key: `/api/v2/workloads/{workload_id}`, snake_case, no trailing slash), so the example returned null and contradicted the file's own prefix table. | Example uses `yq '.paths."/api/v2/workloads/{workload_id}".patch'` — key and method verified present in the docs-site spec. | `references/schema-reference.md` — `f8e661a` |
| Credential-key lookup rule pointed at nonexistent `<Type>Credentials` schemas (`S3Credentials`, `BasicCredentials`, ...) — the documented lookup fails for every unlisted type. | Rule points at the `CredentialsBody` oneOf variant matching the `credentialType` (the `POST /credentials/` body, present in both specs) and warns off the `<Type>CredentialsFields` inline data-source schemas whose field names differ from stored keys. | `references/schema-reference.md` — `f8e661a` |
| Claimed the nonexistent `MultiContainerArtifactSpec` schema covers the artifact `spec` object — the lookup returns null. | Documents `spec` as a discriminated union on `type`: `ServiceArtifactSpec`, `NimArtifactSpec`, `AgentArtifactSpec` (`mcp` maps to `ServiceArtifactSpec`). | `references/schema-reference.md` — `f8e661a` |
| Stated artifacts are created only inline (`CreateWorkloadRequest.artifact`) or by cloning (`ArtifactCloneRequest`) — omitting direct `POST /artifacts/`, contradicting this skill's own code-to-workload flow and hiding the request-body schema from agents. | Adds the direct path: `POST /artifacts/` with request-body schema `InputArtifact`, alongside inline and clone. | `references/schema-reference.md` — `f8e661a` |
| Presented artifact `type` as only `service` (default) or `nim` — the `ArtifactType` enum also accepts `agent` and `mcp`. | Lists all four values: `service` (default), `nim`, `agent`, `mcp`. | `SKILL.md` — `f8e661a` |

**Considered and deliberately not changed:**

- `references/schema-reference.md:51` oauth credential-key row (`token`, `refreshToken`) — REFUTED: the `CredentialsBody` oauth variant (POST /credentials/ request body) stores exactly `token`/`refreshToken`, so the skill's table was already correct; the audit's contrary `OAuthCredentialsFields` evidence belongs to the inline data-source-credential subsystem, and "fixing" it per the audit would have introduced a real error.
- `SKILL.md:169` ```python fence on the `environmentVars` fragment (sdk-discrepancies A9) — DOWNGRADED: the fragment is valid Python-fragment syntax (trailing comma legal in Python, invalid in JSON, so a ```json retag would itself be wrong), matches the skill's httpx `json=` convention, and misleads no agent; pure static-tooling hygiene, left as-is.
- No behavioral round 2-4 follow-ups touched this skill: per-commit diffs for `bb02176`/`11d3c29`/`aa7c685` are empty for `skills/datarobot-workload-api`, and the rounds' "All SDK skills" install-command sweeps did not apply (this skill's only install mention is `pip install httpx`, not a `pip install datarobot` site).

</details>

<details>
<summary><b>datarobot-feature-engineering</b> — 12 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Taught `dr.Model.get("xyz123")` with only a model id in both Pattern code blocks — the real signature is `(project, model_id)`, so every copy-paste run crashed with `TypeError: Model.get() missing 1 required positional argument: 'model_id'` | Both patterns call `dr.Model.get(project_id, model_id)` with a comment noting both ids are required | `SKILL.md` `f8e661a` |
| Quick Start taught an invented helper `get_feature_importance(model_id)` — no such function exists in the SDK or the skill (immediate `NameError`) | Quick Start teaches `dr.Model.get(project_id, model_id)` then `model.get_or_request_feature_impact()` (per the amended verdict, avoiding the uncomputed-impact trap) | `SKILL.md` `f8e661a` |
| Quick Start taught an invented helper `export_feature_list(project_id)` — undefined anywhere (immediate `NameError`) | Teaches `dr.Project.get(project_id).get_features()`, iterating `f.name`, `f.feature_type`, `f.importance` to write the export | `SKILL.md` `f8e661a` |
| Listed nonexistent `model.get_features()` as a real SDK method — `Model` has no such attribute (`AttributeError`), and the nearest real method returns something different from what the line implied | Lists `model.get_features_used()`, documented as returning bare name strings that include the target column | `SKILL.md` `f8e661a` |
| Patterns 1–2 called bare `model.get_feature_impact()`, which raises an unexplained `ClientError` (404) whenever impact was never computed, while the Error-handling section misdirected agents to conclude the model type was unsupported | Patterns use `model.get_or_request_feature_impact()` (computes impact first if needed) and Error handling names the 404 cause and the remedy | `SKILL.md` `f8e661a` |
| Applied fixed `impactNormalized` thresholds without ever defining `impactNormalized`/`impactUnnormalized`, letting agents invent meanings — a behavioral run reported `impactUnnormalized` 0.6519 as "~65% of raw predictive signal" | Defines both fields next to the thresholds: permutation-induced error-metric degradation, normalized so the largest is 1.0 — not a percentage or share of signal | `SKILL.md` `f8e661a` |
| Framed feature impact as revealing "which features drive predictions" without stating scores are unsigned magnitudes or mentioning feature effects — behavioral runs fabricated directional claims presented as model findings | Notes impact is magnitude-only and directs direction questions to partial dependence via `model.get_or_request_feature_effect(source)` with sources from `model.get_feature_effect_metadata()` | `SKILL.md` `f8e661a` |
| Round-2 follow-up: the magnitude-only warning alone did not change behavior — the agent called feature effects zero times and made five unsupported directional claims (up from one in baseline) | Added Pattern 3, a full feature-effects/partial-dependence recipe, plus an explicit output rule: a report that has not computed feature effects must not claim direction | `SKILL.md` `bb02176` |
| Round-2 follow-up: both install sections taught bare `pip install datarobot`, which produced `command not found: pip` and PEP 668 refusals in behavioral runs | Both sites use `python -m pip install datarobot` with a `uv pip install` fallback note for pip-less venvs and PEP 668 systems | `SKILL.md` `bb02176` |
| Round-3 follow-up: presenting `uv pip` as a separate fallback step cost two wasted tool calls per install in pip-less environments | Install command is a single chained `python -m pip install datarobot \|\| uv pip install datarobot` at both sites, resolving the right installer in one call | `SKILL.md` `11d3c29` |
| Round-3 follow-up: the percentage-only wording let share-of-signal claims survive in qualitative form ("these features carry the majority of the model's predictive power") | Output rule now also forbids share/proportion language about impact scores, with or without a number | `SKILL.md` `11d3c29` |
| Round-4 follow-up: agents rewriting the install line (e.g. adding `-r requirements.txt`) kept the ordering but dropped the `\|\|` operator, losing the installer decision to shell syntax | Prose next to the command states the decision directly: if `python -c "import pip"` fails, go straight to `uv pip install` | `SKILL.md` `aa7c685` |

</details>

<details>
<summary><b>datarobot-model-explainability</b> — 12 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Formatted `result.base_value` with `:.6f` and documented it as `float` — the API returns a single-element list, so the script crashed with `TypeError` after every successful SHAP-matrix compute | Script unwraps `base_value[0]` behind an isinstance guard; reference documents `base_value` as `list[float]` with index `[0]` for the scalar baseline | `scripts/compute_shap_matrix.py`, `references/shap_api_reference.md` `f8e661a` |
| Documented `PredictionExplanationsInitialization.delete(project_id, model_id)` as a class-level call — `delete()` is a no-arg instance method, so the call as written raises `TypeError` | Documents `PredictionExplanationsInitialization.get(project_id, model_id).delete()` and notes `.delete()` takes no arguments | `references/xemp_pe_reference.md` `f8e661a` |
| Documented `create_on_training_data(project_id, model_id, ...)` as running "instead of an uploaded dataset", implying no dataset needed — `dataset_id` is a required positional, so the shown call raises `TypeError` | Header shows the required `dataset_id` and notes OTV/time-series projects must also pass `datetime_prediction_partition` (`"0"` or `"holdout"`) | `references/xemp_pe_reference.md` `f8e661a` |
| Named the `.get_rows()` row field `row_index` — the real attribute is `row_id`, so `row.row_index` raises `AttributeError` only after the full (expensive) PE pipeline has run | Row-structure table names the field `row_id` | `references/xemp_pe_reference.md` `f8e661a` |
| Example explanation dict used the key `"featureValue"` — the SDK underscorizes API camelCase, so `entry["featureValue"]` raises `KeyError` (the real key is `feature_value`) | Example dict uses `"feature_value"` | `references/xemp_pe_reference.md` `f8e661a` |
| Decision guide pointed "Need aggregated importance?" at `ShapImpact.compute(entity_id=model_id)`, which returns a non-blocking `StatusCheckJob` (`.shap_impacts` raises `AttributeError`) and silently inherits the `source="training"` default, while every sibling row returns the finished insight | Row uses `ShapImpact.create(entity_id=model_id, source="validation")`, matching the other decision-guide rows | `SKILL.md` `f8e661a` |
| DataSlice example passed a scalar filter value (`"values": 100000`) — the SDK trafaret and server require a list, so the create fails instead of slicing | Example passes `"values": [100000]` | `SKILL.md` `f8e661a` |
| Taught bare `ShapImpact.create(entity_id=model_id)` (Quick Start row and code example) — unlike the other three insight classes it defaults to `source="training"`, so results were silently mislabeled as validation-based and a bare `.get()` (default `"validation"`) finds nothing | Both sites pass `source="validation"` explicitly, with a note explaining that `ShapImpact.create`/`.compute` default to `"training"` while `.get` and the sibling classes default to `"validation"` | `SKILL.md` `f8e661a` |
| Never stated that `ShapImpact` values are mean absolute SHAP — magnitude only — while the Interpreting section covered only signed per-row semantics; behavioral runs shipped fabricated directional claims from impact rankings | Interpreting section states aggregated impact carries no direction and points to signed per-row values (`ShapMatrix`/`ShapPreview`) or `model.get_feature_effect(source=...)` for direction | `SKILL.md` `f8e661a` |
| Called `PredictionExplanationsInitialization.create(...)` fire-and-forget and proceeded straight to `PredictionExplanations.create()` — `.create()` returns an async `Job`, so the documented sequence can race an unfinished initialization | Both sites capture the returned job and call `init_job.wait_for_completion()` before continuing | `SKILL.md`, `references/xemp_pe_reference.md` `f8e661a` |
| ShapImpact attribute table documented `row_count` as a live `int` — deprecated since SDK v3.6, it returns `None` with a `DeprecationWarning`, silently feeding `None` into reports | Removed the `row_count` row from the attribute table | `references/shap_api_reference.md` `f8e661a` |
| Documented ShapImpact `capping` as `bool` ("whether extreme SHAP values were capped") — it is actually an optional dict of cap bounds, never a bool | Documents `capping` as a dict with `left`/`right` keys, or `None` when no capping applies | `references/shap_api_reference.md` `f8e661a` |

</details>

<details>
<summary><b>datarobot-model-training</b> — 12 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Taught `dr.Model.get(model_id)` in Key SDK Operations — the real signature is `(project, model_id)`, so the call crashes with TypeError | Documents `dr.Model.get(project_id, model_id)` | `SKILL.md` (`f8e661a`) |
| Passed a nonexistent `name=` kwarg to `dr.Dataset.create_from_file(...)` — TypeError before any upload, in Pattern 1 and in the helper script this skill's documented command (SKILL.md:141) invokes | Uploads without a name, then renames via `dataset.modify(name=...)` (verified to PATCH and update `dataset.name` in place) | `SKILL.md`, `skills/datarobot-data-preparation/scripts/upload_dataset.py` (`f8e661a`) |
| Pattern 2 passed the dataset name as a second positional to `create_from_file` — it binds to the `filelike` parameter and `assert_single_parameter` raises TypeError immediately | Uploads with the file path only, then `dataset.modify(name=...)` | `SKILL.md` (`f8e661a`) |
| Referenced nonexistent `dr.PARTITIONING_METHOD.DATETIME` — AttributeError before any API call is made | Passes a `dr.DatetimePartitioningSpecification` instance as `partitioning_method=` | `SKILL.md` (`f8e661a`) |
| Passed `datetime_partition_column`, `multiseries_id_columns`, and `forecast_window_start`/`_end` as direct target-setting kwargs — none exist on `set_target` or `analyze_and_model` | Builds all four into the `DatetimePartitioningSpecification` constructor, with the required `use_time_series=True` | `SKILL.md` (`f8e661a`) |
| Taught `project.start(autopilot_on=True, max_wait=7200)` — `Project.start` is a classmethod requiring `sourcedata` with no `max_wait` (TypeError), and would create a new project rather than start this one | Call removed (with its orphaned comment) — `analyze_and_model(..., max_wait=7200)` already sets the target and starts Autopilot | `SKILL.md` (`f8e661a`) |
| Taught nonexistent `project.wait_for_completion()` — AttributeError; the real method is `wait_for_autopilot` | Uses `project.wait_for_autopilot()`, matching the skill's own correct usage at Pattern 1 and Key SDK Operations | `SKILL.md` (`f8e661a`) |
| Configured time series via `project.set_target(...)`, deprecated since v3.0 and slated for removal in v4.0 — contradicting Pattern 1's own comment that `analyze_and_model` replaces it | Pattern 2 rewritten on `project.analyze_and_model(target=..., mode=..., partitioning_method=spec, max_wait=7200)` | `SKILL.md` (`f8e661a`) |
| Never mentioned that `Dataset.create_from_file` and `project.analyze_and_model` block for up to `max_wait` seconds (default 600) — the "Timeout errors" troubleshooting bullet was unactionable | The bullet now names both calls and says to pass a larger `max_wait` for big datasets or slow environments | `SKILL.md` (`f8e661a`) |
| Taught bare `pip install datarobot` (2 sites) — produced `command not found: pip` and PEP 668 refusals in behavioral runs | `python -m pip install datarobot` with a `uv pip install` fallback note for pip-less venvs and PEP 668 systems | `SKILL.md` (`bb02176`) |
| Presented the `uv pip` fallback as a separate step to try after a failure — pip-less environments cost two wasted failing install calls per run (round 3: 3 of 4 install runs) | Single chained one-liner `python -m pip install datarobot \|\| uv pip install datarobot` that resolves the right installer in one call | `SKILL.md` (`11d3c29`) |
| Kept the installer decision only in shell syntax — round 4 showed agents rewriting the one-liner (e.g. adding `-r requirements.txt`) preserved the ordering but dropped the `\|\|` chain, losing the fallback | States the decision in prose next to the command: if `python -c "import pip"` fails, go straight to `uv pip install` | `SKILL.md` (`aa7c685`) |

**Considered and deliberately not changed:**

- Quick Start entries `upload_dataset(...)`/`create_project(...)`/`start_automl(...)` (SKILL.md:15-17) — deliberately not flagged or changed: helper-script shorthand backed by the Helper Scripts section, not literal SDK calls (unlike the literal Key SDK Operations list).
- `dr.ModelRecommendation.get(project.id).get_model()` chaining (SKILL.md:117/:204) — left alone: the `Optional` return only arises when `recommendation_type` is passed; the skill's default-path usage is safe.
- behavioral-run-errors.md's claim that the positional `create_from_file` form "fails silently and later" — refuted by live repro (it raises TypeError immediately via `assert_single_parameter`); the call site was still a defect and was fixed, only the failure-mode characterization was rejected.
- behavioral-run-errors.md line 96 listing SKILL.md:228 under the `name=` kwarg defect — mislabeled (that site is the distinct positional-`filelike` form); both sites were fixed as their actual defects, so no extra change was needed.

</details>

<details>
<summary><b>datarobot-model-monitoring</b> — 11 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Recommended `deployment.get_target_drift()` (Quick Start + SDK operations) with no warning that SDK 3.18.0's `TargetDrift` deserializer requires `segment_value` — it crashes with `DataError: {'segment_value': 'is required'}` on any unsegmented deployment, the common case | Warns about the SDK bug at the bullet and gives the live-verified raw-REST workaround: `dr.Client().get(f"deployments/{deployment_id}/targetDrift/").json()`, reading camelCase keys `driftScore`/`targetName`/`sampleSize` | `SKILL.md` `f8e661a` |
| Taught `model.get_roc_curve()` with no arguments — the real signature requires `source`, so the copied call raises `TypeError: missing a required argument: 'source'` | Shows `model.get_roc_curve("validation")` and lists the valid sources (`"validation"`, `"crossValidation"`, `"holdout"`) | `SKILL.md` `f8e661a` |
| Pattern 1 labeled `stats.metrics['responseTime']` "Mean response time (ms)" — the value is the p50 quantile by default, so agent health reports carried a guaranteed-wrong statistical label (live: p50=243ms vs p90=609ms) | Labels it "Median response time (ms)" and notes it is p50 by default, adjustable via `response_time_quantile` | `SKILL.md` `f8e661a` |
| Invited time-windowed monitoring calls without mentioning that the endpoints reject `start_time`/`end_time` not aligned to the top of an hour — a natural `datetime.now()` window earns HTTP 400 "datetime that is not at the top of an hour is not supported" | Documents the hour-alignment requirement next to the monitoring bullets with a truncation snippet | `SKILL.md` `f8e661a` |
| Quick Start recommended `get_service_stats_over_time(...)` without ever showing the bucket shape — agents guessed keys like `prediction_count` and hit KeyError or, via `.get(..., 0)`, silently summed 0 | Documents that each element of `.buckets` is `{period, model_id, value}` with the metric under `"value"`, e.g. `total = sum(b["value"] or 0 for b in sot.buckets)` (verified live) | `SKILL.md` `f8e661a` |
| The audit's own hour-alignment snippet truncated `end_time` *backward* (with deprecated naive `datetime.utcnow()`), silently dropping the in-progress hour's traffic — a round-2 volume report confidently answered 0 predictions for a window that had served 60 | Truncates forward to the top of the next hour with timezone-aware `(datetime.now(timezone.utc) + timedelta(hours=1)).replace(...)`, explains why back-truncation is wrong, and notes the API accepts a future `end_time` | `SKILL.md` `bb02176` |
| The raw `targetDrift` REST workaround showed no time window, and the natural param guesses `startTime`/`endTime` are rejected with 400 — round 2's only remaining friction signal | Documents `params={"start": ..., "end": ...}` (optional, hour-aligned ISO-8601, default last 7 days) and warns that `startTime`/`endTime` 400 | `SKILL.md` `bb02176` |
| Taught bare `pip install datarobot` (both install sections), which produced `command not found: pip` and PEP 668 refusals in behavioral runs | Uses interpreter-anchored `python -m pip install datarobot` with a `uv pip install` fallback note for pip-less venvs and PEP 668 systems | `SKILL.md` `bb02176` |
| Presented the `uv pip` fallback as a separate follow-up step, costing two wasted calls per install in pip-less environments | Chains the install in one command — `python -m pip install datarobot \|\| uv pip install datarobot` — at both sites so a single call resolves the right installer | `SKILL.md` `11d3c29` |
| Pattern 2 filtered feature drift at an undocumented `> 0.2` threshold that contradicted the skill's own 0.15/0.3 alert-threshold table | Filters at `> 0.15` with a comment pointing at the documented medium/high thresholds | `SKILL.md` `11d3c29` |
| The installer decision lived only in the `\|\|` shell operator — round 4 showed agents that rewrite the command (e.g. adding `-r requirements.txt`) keep the ordering but drop the operator | States the decision in prose next to both install commands: when `python -c "import pip"` fails, go straight to `uv pip install` | `SKILL.md` `aa7c685` |

</details>

<details>
<summary><b>datarobot-model-deployment</b> — 10 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Taught deployment creation with no prediction-server argument anywhere (both creation params are typed `Optional`) — but DataRobot Cloud requires `defaultPredictionServerId`, so the documented flow 422'd only after the training wait was already paid | Key SDK Operations and both patterns now call `dr.PredictionServer.list()` and pass `default_prediction_server_id=servers[0].id`, state the Cloud 422 despite the `Optional` typing, note `PredictionServer` exposes only `.id`/`.url`/`.datarobot_key` (no `.name`), and give the serverless `prediction_environment_id` alternative | `SKILL.md` `f8e661a` |
| Ranked models with `max(models, key=lambda m: m.metrics.get("AUC", 0))` — `model.metrics` values are per-partition dicts, so classification projects crashed with TypeError and regression projects (no AUC key) silently deployed an arbitrary model as "best" | Quick Start and Pattern 1 get the champion via `dr.ModelRecommendation.get(project_id).get_model()` | `SKILL.md` `f8e661a` |
| Taught `dr.Deployment.list(project_id)` — the method has no `project_id` parameter, the positional binds to `order_by`, and the call 400s | Teaches `dr.Deployment.list()` (optionally `search=<label substring>`) and states explicitly that deployments cannot be filtered by project id via the SDK | `SKILL.md` `f8e661a` |
| Taught `deployment.perform_model_replace(new_model_id=..., reason=...)` — the real kwarg is `new_registered_model_version_id`, so the call raised TypeError, and merely renaming the kwarg still fails server-side because a leaderboard model id is not a registered model version id | Documents the full flow: register the new model via `create_for_leaderboard_item(model_id=..., registered_model_id=<deployed registered model>)`, then `perform_model_replace(new_registered_model_version_id=version.id, reason=...)` with the `dr.enums.MODEL_REPLACEMENT_REASON` values enumerated | `SKILL.md` `f8e661a` |
| Quick Start step 2 taught an invented `create_deployment(model_id, deployment_name)` helper that exists nowhere in the SDK or the skill (no `scripts/` dir), crashing on first use | Quick Start walks the real register-then-deploy flow: `dr.RegisteredModelVersion.create_for_leaderboard_item(...)` then `dr.Deployment.create_from_registered_model_version(model_package_id=version.id, label=..., default_prediction_server_id=...)` | `SKILL.md` `f8e661a` |
| Quick Start step 3 taught an invented `get_deployment_endpoint(deployment_id)` — "retrieve deployment endpoints" was a listed capability yet no real SDK method for it appeared anywhere in the skill | Teaches `(deployment.default_prediction_server or {}).get("url")` (returns `None` for DataRobot Serverless deployments) and documents the dict's kebab-cased `datarobot-key` header key | `SKILL.md` `f8e661a` |
| Taught deprecated `dr.Deployment.create_from_learning_model` at three sites (deprecated since v3.8, removal scheduled v3.12 — six minor versions behind the installed 3.18.0, surviving only because removal slipped) | All three sites use the registered-model flow (`create_for_leaderboard_item` → `create_from_registered_model_version`), and Key SDK Operations explicitly says do not use the deprecated method | `SKILL.md` `f8e661a` |
| Install instructions said bare `pip install datarobot` at two sites, which produced `command not found: pip` and PEP 668 refusals in behavioral runs | Both sites use `python -m pip install datarobot` with a `uv pip install datarobot` fallback note for pip-less venvs and PEP 668 systems | `SKILL.md` `bb02176` |
| Presenting the `uv pip` fallback as a separate manual step cost two wasted tool calls per install in pip-less environments (behavioral round 3) | Both install sites use the single chained command `python -m pip install datarobot \|\| uv pip install datarobot`, which resolves the right installer in one call | `SKILL.md` `11d3c29` |
| Behavioral round 4 showed agents that rewrite the install command (e.g. adding `-r requirements.txt`) keep the ordering but drop the `\|\|` operator, silently losing the fallback | The installer decision now survives outside shell syntax: prose next to both commands says if `python -c "import pip"` fails, go straight to `uv pip install` | `SKILL.md` `aa7c685` |

</details>

<details>
<summary><b>datarobot-external-agent-monitoring</b> — 9 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Taught `from opentelemetry.instrumentation.httpx import HTTPXInstrumentor` — that class does not exist in opentelemetry-instrumentation-httpx, so the copy-paste block crashed with ImportError | Imports and calls the real `HTTPXClientInstrumentor().instrument()` | `frameworks/generic-python.md` `f8e661a` |
| ADK `configure_otel()` template passed `preferred_temporality=preferred_temporality` without ever defining the name (the elision marker covered imports only), so the copied template raised NameError at the app's first import | Template defines the `{Counter/Histogram/ObservableCounter: AggregationTemporality.DELTA}` dict before the `OTLPMetricExporter` call | `frameworks/google-adk.md` `f8e661a` |
| Taught a "built-in" fallback import `llama_index.core.callbacks.open_inference_callback` that has never existed in llama-index-core — ModuleNotFoundError, with the "built-in" framing sending agents hunting inside core | Imports `OpenInferenceCallbackHandler` from the real path `llama_index.callbacks.openinference`, states the required `pip install llama-index-callbacks-openinference`, and drops the "built-in" framing | `frameworks/llamaindex.md` `f8e661a` |
| Step 2's only install was bare `pip install datarobot`, yet Step 5's `verify_otel_connection.py` imports `opentelemetry.*` (ModuleNotFoundError — neither `datarobot` nor `datarobot[otel]` pulls the OTLP HTTP exporter) and bare `pip` fails in pip-less uv venvs and under PEP 668; the script-environment Dependencies block also omitted the OTel packages | Install step is interpreter-anchored `python -m pip install` covering `datarobot` plus the three OTel packages the verify script imports, with a `uv pip install` fallback, and the script-environment dependency block now lists those OTel packages | `SKILL.md` `f8e661a` |
| Gave the optional Logfire dependency as `logfire[pydantic-ai]` — logfire has no such extra (installers emit a warning and pull nothing additional) | Plain `logfire`; the base package already provides `instrument_pydantic_ai()` and `send_to_logfire` | `frameworks/pydantic-ai.md` `f8e661a` |
| Documented `InstrumentationSettings(version=…)` as `Literal[2, 3, 4, 5]` and "every selectable format, 2–5" — current pydantic-ai accepts `Literal[2, 3, 4, 5, 6]` | Documents `Literal[2, 3, 4, 5, 6]` (default 5) and "2–6", keeping the still-true note that no version emits `gen_ai.prompt`/`gen_ai.completion` | `frameworks/pydantic-ai.md` `f8e661a` |
| Step 4 said to add the OTel packages but never to install the target project's own declared dependencies, so instrumenting a project the agent did not set up crashed after successful instrumentation (round 3's only scenario failure: `ModuleNotFoundError: langchain_core` at +186s of a 231s run) | Step 4 says to install the project's full dependency set (`-r requirements.txt`, `uv sync`, or `poetry install`) before running it | `SKILL.md` `11d3c29` |
| Presented the `uv pip` fallback as a separate follow-up step after a failed `python -m pip`, costing two wasted tool calls per install in pip-less environments | Install command is one chained `python -m pip install … \|\| uv pip install …` that resolves the right installer in a single call | `SKILL.md` `11d3c29` |
| Encoded the installer decision only in shell syntax — round 4 showed an agent rewriting the command (adding `-r requirements.txt`) keeps the ordering but drops the `\|\|` operator, burning a call on the pip-less path | Prose beside the command states the decision itself: if `python -c "import pip"` fails, go straight to `uv pip install` | `SKILL.md` `aa7c685` |

</details>

<details>
<summary><b>datarobot-data-preparation</b> — 8 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Taught `dr.Dataset.create_from_file(file_path, name)` — the real signature has no `name` parameter, so the documented pattern and the helper script crashed with `TypeError` before any upload | Uploads with `create_from_file(file_path, ...)` then renames via `dataset.modify(name=...)`; the script keeps its CLI unchanged | `SKILL.md` (SDK reference + Pattern 1), `scripts/upload_dataset.py` — `f8e661a` |
| Read nonexistent `dataset.column_count` (4 sites) — `AttributeError` on every access; in the script it struck after the upload succeeded, stranding the created dataset without ever printing its ID | Computes column count as `len(dataset.get_all_features())`, which exhausts pagination so the count is the true total | `SKILL.md` (3 sites), `scripts/upload_dataset.py` — `f8e661a` |
| Quick Start step 2 told agents to call `validate_dataset(dataset_id)` — no such helper script or SDK method exists, and step 1's identical-looking helper is real, so agents wasted turns hunting for it | Teaches a concrete quality-check recipe: iterate `dataset.get_all_features()` and inspect each feature's `na_count`, `unique_count`, and `target_leakage` | `SKILL.md` — `f8e661a` |
| Quick Start step 3 told agents to call nonexistent `get_dataset_schema(dataset_id)` to review structure | Teaches `dr.Dataset.get(dataset_id).get_all_features()`, reading each column's `name` and `feature_type` | `SKILL.md` — `f8e661a` |
| Never mentioned that `create_from_file` blocks until server-side ingestion completes (default `max_wait=600`s), so agents ran uploads under 120s foreground shell timeouts and were killed mid-ingestion | Notes the blocking behavior beside the SDK reference, in Pattern 1, and in the script docstring: allow several minutes or run in background, and raise `max_wait` for large files | `SKILL.md`, `scripts/upload_dataset.py` — `f8e661a` |
| Both install sections said bare `pip install datarobot`, which produced `command not found: pip` and PEP 668 refusals in behavioral runs (round 2) | Uses `python -m pip install datarobot` plus a `uv pip install` fallback note for pip-less venvs and PEP 668 systems | `SKILL.md` (2 sites) — `bb02176` |
| The `uv pip` fallback as a separate prose step cost two wasted shell calls per install in pip-less environments — `pip` and `python -m pip` both fail before the agent reaches `uv` (round 3) | Install command is a single chain, `python -m pip install datarobot \|\| uv pip install datarobot`, resolving the right installer in one call | `SKILL.md` (2 sites) — `11d3c29` |
| The `\|\|`-encoded fallback did not survive command rewrites — agents adding extra packages or `-r requirements.txt` kept the ordering but dropped the operator, losing the fallback (round 4) | States the installer decision in prose beside the command: when `python -c "import pip"` fails, go straight to `uv pip install` | `SKILL.md` (2 sites) — `aa7c685` |

</details>

<details>
<summary><b>datarobot-app-framework-cicd</b> — 4 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Simple Path CD and PR-preview workflows exported only `DATAROBOT_API_TOKEN` and `PULUMI_ACCESS_TOKEN` — SDK 3.18 has no default endpoint, so the reference template's `pulumi up` (top-level `dr.client.get_client()` via its feature-flag check) crashed with `ValueError: No valid configuration found` on every fresh runner | Both env blocks now set `DATAROBOT_ENDPOINT: ${{ vars.DATAROBOT_ENDPOINT \|\| 'https://app.datarobot.com/api/v2' }}` (value live-verified) and the header comments document the optional variable | `examples/github-cd-pulumi-cloud.yml`, `examples/github-deploy-pulumi-cloud.yml` — `f8e661a` |
| Destroy workflow had the same `DATAROBOT_ENDPOINT` omission — downgraded on verification (`pulumi destroy` is state-only and the bridged provider falls back to US SaaS, so token-only destroy worked in the default case) but it silently targeted the wrong host for EU/on-prem installs and dynamic-resource deletes | Same `${{ vars.DATAROBOT_ENDPOINT \|\| 'https://app.datarobot.com/api/v2' }}` env line and header note added for consistency with the other two workflows and EU/on-prem correctness | `examples/github-destroy-pulumi-cloud.yml` — `f8e661a` |
| Simple Path docs declared the two GitHub Secrets the complete credential set ("both required — no defaults") and never mentioned `DATAROBOT_ENDPOINT`, so agents following the skill's default path shipped pipelines that failed on first deploy | Optional GitHub Variables table now lists `DATAROBOT_ENDPOINT` (default `https://app.datarobot.com/api/v2`) with a note to set it only for EU or on-prem/self-managed installs | `SKILL.md` — `f8e661a` |
| The setup guide the skill copies into user repos asserted "Both secrets are required. The workflows will not run without them." as the full one-time setup, omitting the endpoint entirely | Step 4 now covers repository Variables (added on the Variables tab per the amended fix, not as a Secrets-table row) including `DATAROBOT_ENDPOINT` with the US default and an EU example | `examples/workflows-README.md` — `f8e661a` |

</details>

<details>
<summary><b>datarobot-discover</b> — 2 defects fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Taught `npx ai-agent-skills install <skill-name>` as the sole install command, calling ai-agent-skills "the DataRobot skills plugin" — bare names resolve against that third-party installer's own catalog, which contains zero DataRobot skills, so the documented install fails deterministically for every skill | Teaches the repo-path form `npx ai-agent-skills install datarobot-oss/datarobot-agent-skills/skills/<skill-name>` (plus the whole-repo form for all skills) and describes ai-agent-skills as a third-party installer pointed at the DataRobot repo | `SKILL.md` `f8e661a` |
| Told the agent to display each catalog entry's **name** field — live `/.well-known/ai-catalog.json` entries carry no `name` key (only `displayName`), so a scripted lookup like `jq '.[].name'` silently prints null for every entry | Lists entries by **displayName**, the key verified present on all live catalog entries | `SKILL.md` `f8e661a` |

</details>

<details>
<summary><b>datarobot-agent-assist</b> — 1 defect fixed</summary>

| Defect | Fix | Where |
|---|---|---|
| Dress Rehearsal Step 1 taught `rehearsal.py --init [--spec agent_spec.md]` without the required `--target-dir` — the command errors out (`Error: --target-dir is required for --init`), and the standalone-triggerable sub-skill never defines `<target_dir>`, so an agent must guess it to recover | Teaches the verified form `rehearsal.py --init --spec <target_dir>/agent_spec.md --target-dir <target_dir>`, matching the canonical `references/dress-rehearsal.md` invocation | `agent-assist-build/SKILL.md` `f8e661a` |

</details>

---

`datarobot-setup` was audited and found clean — no changes.

---

> [!NOTE]
> **Low Risk**
> Changes are documentation, example workflows, and skill helper scripts—no production services or auth/data-plane logic. Residual risk is agents following updated guidance incorrectly in customer environments, not runtime regressions in this repo.
> 
> **Overview**
> **Release 1.6.1** aligns plugin manifests (`package.json`, Claude/Cursor/Gemini) and documents a repo-wide repair of agent skills against **DataRobot Python SDK 3.18.0**, validated with live read-only API checks and four rounds of behavioral agent runs.
> 
> The bulk of the diff **corrects SKILL.md patterns and helper scripts** so documented calls match the real SDK and API: registered-model deployment instead of deprecated `create_from_learning_model`, two-arg `Model.get`, `deployment.get_features()` dicts, upload-then-`modify(name=...)`, time series via `DatetimePartitioningSpecification` + `analyze_and_model`, batch scoring via `BatchPredictionJob.score(deployment=...)`, and workload-api diagnose/telemetry shapes (OTel error logs, `eventType`, docs-site OpenAPI, autoscaling field names).
> 
> **Cross-cutting agent UX** changes standardize SDK install as `python -m pip … || uv pip …` with prose fallback when agents rewrite commands, set **`DATAROBOT_ENDPOINT`** in App Framework CI workflow examples (SDK 3.x has no default), fix discover install paths and catalog **`displayName`**, and tighten explainability/monitoring/feature-engineering guidance so agents do not emit wrong drift windows, directional claims without feature effects, or invalid categorical placeholders (prediction templates now pull training levels/medians from `dr.Feature`).
> 
> **`CHANGELOG.md`** captures per-skill fix lists and behavioral follow-ups; **`datarobot-setup`** unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0dc1df417d960fdc59dbd41d34633e9738183e9a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


skills/datarobot-workload-api/scripts/wait_for_running.py
```diff
@@ -49,7 +48,7 @@ def wait_for_running(
         if status in TERMINAL_FAILURES:
             raise RuntimeError(
                 f"Workload {workload_id} entered terminal state {status!r}. "
-                f"statusDetails={json.dumps(w.get('statusDetails'), default=str)[:500]}"
+                f"Run: python scripts/diagnose_workload.py {workload_id}"
```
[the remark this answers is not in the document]
mhauskn-dr: Valid finding — fixed in 0dc1df4. The message now resolves the diagnose script via `Path(__file__).parent / "diagnose_workload.py"`, matching the sibling `wait_for_replacement.py`. Good catch that this fires at exactly the moment an operator/agent is recovering from a terminal failure, where a non-resolving path costs the most.

skills/datarobot-predictions/scripts/validate_prediction_data.py
```diff
@@ -89,24 +93,16 @@ def validate_prediction_data(deployment_id: str, file_path: str) -> dict:
...
-            )
-    except dr.errors.ClientError as e:
-        print(
-            f"Note: feature impact unavailable, skipping importance warnings: {e}",
-            file=sys.stderr,
+    # Use the importance already returned by deployment.get_features() for warnings
+    # (a model-independent measure of feature/target relationship strength)
+    low_importance_features = [
+        f["name"] for f in features if (f["importance"] or 0) < 0.05
+    ]
```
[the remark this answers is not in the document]
mhauskn-dr: False positive — the premise doesn't hold for this field. `deployment.get_features()` `importance` is not a signed score: the SDK docstring defines it as a "numeric measure of the relationship **strength** between the feature and target (independent of model or other features)". Direction is not encoded, so an anti-correlated feature registers as *positive* strength — "strongly negative = important anti-correlated column" is not a state this field can express. Negative values are near-zero artifacts meaning "no detectable marginal relationship" (every negative observed live on app.datarobot.com was tiny: −0.0009, −0.0777), which is exactly what the low-importance warning should capture. Taking `abs()` would mislabel those noise features as important.

Left the logic as-is and added a comment in 0dc1df4 stating the signed comparison is intentional, so this doesn't get "fixed" in a future pass.

mhauskn-dr:
bugbot run

skills/datarobot-data-preparation/scripts/upload_dataset.py
```diff
@@ -10,6 +10,10 @@
 
 Supports CSV, Parquet, and other formats. Optionally links the dataset to
 an existing Use Case so it isn't orphaned in the DataRobot UI.
+
+Note: the upload is a blocking call that returns only after server-side
+ingestion completes (default max_wait=600 seconds) — run it with an adequate
+timeout or in the background, and raise max_wait for large files.
```
damon-sava-stanley: What does "large files" mean here? Or "adequate" timeout? Hard to say in advance/in general (it depends on the vagaries of your internet connection among other thing). And maybe being more prescriptive is out of scope for this PR, but I suspect it mgiht be better to say nothing here.

skills/datarobot-data-preparation/SKILL.md
```diff
@@ -12,8 +12,8 @@ This skill provides guidance for preparing and managing data in DataRobot, inclu
 **Most common use case**: Upload and validate a dataset
 
 1. **Upload dataset**: `upload_dataset(file_path, dataset_name)` to upload data
-2. **Validate data**: `validate_dataset(dataset_id)` to check data quality
-3. **Check schema**: `get_dataset_schema(dataset_id)` to review structure
+2. **Validate data**: iterate `dataset.get_all_features()` and check each feature's `na_count`, `unique_count`, and `target_leakage` for quality issues
```
damon-sava-stanley: Does the skill say what quality issues to look for? Also 2/3 now look to be identical.

chasdr (commented):
ran an independent verification pass rather than just reading the diff:

- checked ~25 of the SDK claims against a clean `datarobot==3.18.0` install via `inspect.signature`: `create_from_file` really has no `name` param (second positional is `filelike`, exactly the bug described), no `Dataset.column_count`, two-arg `Model.get`, no `Model.get_features`, `get_roc_curve(source)` required, `Deployment.list()` without `project_id`, `perform_model_replace(new_registered_model_version_id=...)`, `PredictionServer` exposing only `id`/`url`/`datarobot_key`, `BatchPredictionJob.score(deployment, ...)`, `Project.start` being a classmethod, `PredictionExplanationsInitialization.create` returning a `Job`, `create_on_training_data` requiring `dataset_id`, the `MODEL_REPLACEMENT_REASON` values, and the `ShapImpact` create=training / get=validation default asymmetry. every one matched.
- ran both discover install commands live in a sandboxed HOME: the repo-path form installs `datarobot-discover`, the bare repo path installs all 14 skills. the bare-name failure claim is consistent with the installer resolving bare names against its own upstream catalog only.
- grepped for sweep leftovers: no stale `1.6.0` in any of the five manifests, no leftover `logTail`/`minCount`/`updating` references in the workload-api docs, imports match usage in the touched scripts.
- the `deployment.get_features()` importance rebuttal to bugbot checks out against the SDK docstring (model-independent relationship strength, can be negative, so the signed `< 0.05` comparison is right).
- the google-adk `preferred_temporality` addition fixes a real NameError: the exporter call below already referenced the undefined dict, and the generic pattern carries the imports.
- 305 tests pass locally; my one local failure is the `gemini` binary missing on my machine, and CI has it green.

two minor inline comments, nothing blocking.

skills/datarobot-predictions/scripts/generate_prediction_data_template.py
```diff
@@ -34,45 +34,88 @@ def generate_prediction_data_template(
...
+        for feature in prediction_features:
+            try:
+                if feature["feature_type"] == "Categorical":
+                    histogram = dr.Feature.get(
+                        project_id, feature["name"]
+                    ).get_histogram()
+                    # Each bin is {"label": <level>, "count": ..., "target": ...};
+                    # high-cardinality features get at most 60 bins by default.
+                    categorical_levels[feature["name"]] = [
+                        b["label"] for b in histogram.plot
```
chasdr: not my area so i may be off here, but don't the histogram bins for high-cardinality categoricals include the aggregate pseudo-buckets (`=All Other=`, and `==Missing==` when there are NAs)? if so those literals get cycled into the template rows and the "Valid values" header as if they were real levels.

maybe filter labels matching the `=...=` pattern before use?

CHANGELOG.md
```diff
@@ -11,6 +11,43 @@ Each entry should be prefixed with the affected skill folder name (for example,
 
 ## [Unreleased]
 
+## [1.6.1] - 2026-08-27
```
chasdr: the PR body still says the bump is 1.5.6, and this date is 08-27. worth syncing both to what actually lands at merge?

chasdr:
i think this is ok to merge, but with 42 files across 12 skills and most of them outside my area, i'm not comfortable being the only signoff. @carsongee @andriykislitsyn @elvinagam can you take a pass before this lands?

skills/datarobot-predictions/scripts/generate_prediction_data_template.py
mhauskn-dr: Confirmed and fixed in b339f0c — verified live first: =All Other= appears for any categorical past the bin cap and sorts first for letter-valued features (so it landed in row 1), and ==Missing== appears at any cardinality when training had NAs. Labels wrapped in = are now filtered, and the header carries what they signify instead: "(top 36 of 40 levels)", "(missing allowed - leave blank)".

Since this was a data-shape gap the test fixture couldn't catch, we swept the whole class rather than just this instance: every helper script and runnable doc snippet in the PR, executed read-only against all 79 deployments, 7 workloads, and every project in a live org. That surfaced 8 more same-class defects — fixed in 9237b05 (worst: a crash on deployments whose training project was deleted, 30/79 here; the generate→validate round-trip failing on its own metadata lines; OTel trace/span units off by 1000×). Post-fix the sweep runs clean: 0 tracebacks across 237 script invocations, and all 41 generated templates validate round-trip (previously 0). Good catch — it paid for itself several times over.

CHANGELOG.md
mhauskn-dr: Both right. The changelog date is bumped (b339f0c). The PR body's version line stays as-is deliberately until approvals are in — we'll sync it to whatever actually lands (1.6.1 as of the current main merge) as the final pre-merge step, so it gets corrected exactly once.

.claude-plugin/marketplace.json
```diff
@@ -8,7 +8,7 @@
       "name": "datarobot-agent-skills",
       "source": "./",
       "description": "DataRobot skills for AI/ML workflows — model training, deployment, predictions, feature engineering, monitoring, explainability, data preparation, App Framework CI/CD, and external agent monitoring.",
-      "version": "1.6.0"
+      "version": "1.6.1"
```
carsongee: This should be 1.7.0. There's quite a few changes here, and we're likely to want to patch 1.7.0 if some of these changes need fixes down the road

skills/datarobot-feature-engineering/SKILL.md
```diff
@@ -197,9 +237,13 @@ Common errors and solutions:
 ### Install DataRobot SDK
 
 ```bash
-pip install datarobot
+python -m pip install datarobot || uv pip install datarobot
```
carsongee: This is backwards. You'd want uv to go first for better performance and package caching

skills/datarobot-model-deployment/SKILL.md
```diff
@@ -89,29 +90,36 @@ Use this skill when you need to:
 This skill guides you to use the DataRobot Python SDK directly. Install the SDK if needed:
 
 ```bash
-pip install datarobot
+python -m pip install datarobot || uv pip install datarobot
```
carsongee: Backward here too. Also, the extra `python -m pip` is just a result of masking Python pathing issues over using `pip`. I'd prefer we keep pip for A. Simpler instruction and B. Lower token count of the skill

skills/datarobot-model-training/SKILL.md
```diff
@@ -91,9 +91,13 @@ Use this skill when you need to:
...
 
 ```bash
-pip install datarobot
+python -m pip install datarobot || uv pip install datarobot
 ```
 
+The `uv pip` fallback covers environments with no pip (uv-created venvs, PEP 668 systems). If you
+rewrite this command (extra packages, `-r requirements.txt`), decide the installer first: when
+`python -c "import pip"` fails, go straight to `uv pip install`.
+
```
carsongee: Yeah, same here. Lots of extra context. Not very useful since the model will figure that based on whatever error messages happen. Skills that explain their own "why" tend to just waste context and tokens

skills/datarobot-model-training/SKILL.md
```diff
@@ -271,16 +278,20 @@ Common errors and solutions:
...
```
carsongee: Wow, it really believes in this. That suggests something could be wrong with the harness you ran the tests and validation in more than the actual skill or the results of running the automated improvements

skills/datarobot-predictions/scripts/generate_prediction_data_template.py
```diff
@@ -34,45 +35,134 @@ def generate_prediction_data_template(
...
+                elif feature["feature_type"] in ("Numeric", "Percentage"):
+                    stats = dr.Feature.get(project_id, feature["name"])
+                    if isinstance(stats.median, (int, float)):
+                        numeric_stats[feature["name"]] = {
+                            "min": stats.min,
+                            "max": stats.max,
+                            "median": stats.median,
+                        }
+            except Exception:
+                pass  # keep the type placeholder for this feature
```
carsongee, line 112: That's a whole lot more code. @danna-naser this seems like more of an opportunity to improve our SDK than the skill. But of course, I'll leave that your expertise

skills/datarobot-predictions/scripts/validate_prediction_data.py
```diff
@@ -89,24 +101,21 @@ def validate_prediction_data(deployment_id: str, file_path: str) -> dict:
...
-            )
-    except dr.errors.ClientError as e:
-        print(
-            f"Note: feature impact unavailable, skipping importance warnings: {e}",
-            file=sys.stderr,
+    # Use the importance already returned by deployment.get_features() for warnings
+    # (a model-independent measure of feature/target relationship strength; direction
+    # is not encoded — negative values mean no detectable relationship, so the signed
+    # comparison below is intentional: don't take abs()). None means importance is
+    # unknown for this model class (unsupervised, time-series, custom) — not low.
```
carsongee, line 108: Who is this comment for? Neither human or LLM will read it unless this code is broken

skills/datarobot-predictions/SKILL.md
```diff
@@ -105,16 +105,23 @@ Execute predictions using various methods:
 This skill guides you to use the DataRobot Python SDK directly. Install the SDK if needed:
 
 ```bash
-pip install datarobot datarobot-predict
+python -m pip install datarobot datarobot-predict || uv pip install datarobot datarobot-predict
 ```
 
+The `uv pip` fallback covers environments with no pip (uv-created venvs, PEP 668 systems). If you
+rewrite this command (extra packages, `-r requirements.txt`), decide the installer first: when
+`python -c "import pip"` fails, go straight to `uv pip install`.
```
carsongee: Wow, fun. tons of dupes here...

carsongee (commented):
Very interesting results! For the areas I wrote like the CI/CD, the improvements were minor but real. The only real issue I see is the pip/uv silliness

.claude-plugin/marketplace.json
mhauskn-dr: Superseded by the new auto-versioning — rebased onto main, all version bumps removed

skills/datarobot-feature-engineering/SKILL.md
mhauskn-dr: Done in 0d707e9 — uv pip install … || python -m pip install … at all 14 sites, and the fallback-rationale prose is gone entirely (~250 tokens back).

skills/datarobot-model-deployment/SKILL.md
mhauskn-dr: Adopted uv-first and dropped the prose. Kept python -m pip (rather than bare pip) as the fallback — it anchors the install to the interpreter that runs the code. If you feel strongly about using a bare pip, then happy to update.

skills/datarobot-model-training/SKILL.md
mhauskn-dr: prose removed

skills/datarobot-model-training/SKILL.md
mhauskn-dr: prose removed

skills/datarobot-predictions/scripts/validate_prediction_data.py
mhauskn-dr: It's for automated reviewers, concretely: Cursor Bugbot flagged this exact line as a bug in its first pass on this PR (suggesting abs(), which would mislabel noise features as important). The comment is what stops the next bot or reviewer from "fixing" correct code.

skills/datarobot-predictions/SKILL.md
mhauskn-dr: repeated prose now deleted

PR #100, opened 2026-09-02
[CFX-7986] agent-assist: write the external LLM env the template reads

chasdr:
## What

The agent-assist skill's external LLM path (shipped in 1.8.0) wrote `EXTERNAL_LLM_MODEL/API_KEY/BASE_URL` to the generated app's `.env`, which nothing in the agent template or its shared LLM library reads. Pointing an agent at an OpenAI-compatible endpoint (a local LiteLLM proxy, say) silently failed: the model was never reached, and `task start-non-interactive` died at the deploy stage.

## Changes

- `setup_template.py`: the external path writes what the template actually reads to resolve an external LLM: `USE_DATAROBOT_LLM_GATEWAY=0`, `OPENAI_API_BASE`, `OPENAI_API_KEY`, and an `openai/`-prefixed `LLM_DEFAULT_MODEL` so litellm routes to the base URL.
- It stops after writing `.env` for the external case. `dr dotenv setup` rewrites the file onto the gateway path (dropping the OpenAI keys) and the Pulumi deploy wants a governed DataRobot deployment, neither of which fits a raw base URL. It also writes a generated `SESSION_SECRET_KEY` (the FastAPI server requires it) so `task dev` runs, and rejects a base URL or key carrying `.env`-breaking characters.
- Docs (llm-selection, helper-scripts, agent-spec-schema, pre-coding-checklist, SKILL) and CHANGELOG describe the external source with the correct contract.

## Verification

Built the app's LLM from the produced `.env` and confirmed it routes end to end to a local OpenAI-compatible proxy: external path selected, model name stripped on the wire, client auth accepted. Setup produces the correct `.env` and runs no dotenv or deploy for the external case.

## Scope

Local-run path. `DATAROBOT_ENDPOINT`/`DATAROBOT_API_TOKEN` are left empty, so the LLM works but DataRobot-backed app features (auth, chat history) need them set. Deploying an external-LLM agent is a separate follow-up.

---

> [!NOTE]
> **Low Risk**
> Scoped to agent-assist setup scripting and documentation; gateway and deployed LLM paths are unchanged aside from clearer prereq wording.
> 
> **Overview**
> Fixes **`datarobot-agent-assist`** external OpenAI-compatible LLM setup, which previously wrote unread `EXTERNAL_LLM_*` keys so agents never reached the configured endpoint and still ran DataRobot-only deploy steps.
> 
> **`setup_template.py`** now writes **`USE_DATAROBOT_LLM_GATEWAY=0`**, **`OPENAI_API_BASE`**, **`OPENAI_API_KEY`**, and an **`openai/`-prefixed `LLM_DEFAULT_MODEL`** (from `AGENT_ASSIST_LLM_*` via `--llm-base-url`). It adds **`SESSION_SECRET_KEY`** for local FastAPI boot, validates dangerous characters in URL/key, and **returns after `.env`**—skipping **`dr dotenv setup`** and Pulumi so gateway rewrite/deploy do not break the external path.
> 
> Skill docs, **`llm_base_url`** in the agent spec schema, and **CHANGELOG** document the third **`external`** model source and local **`task dev`** workflow (deploy still out of scope).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7832448a168b56a8c246dcbc780878bc472a957. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Version bumps, `[Unreleased]` renames, and releases are automated&mdash;see
 
 ## [Unreleased]
 
+### Fixed
+
+- `datarobot-agent-assist`: External LLM setup writes the OpenAI-compatible env the agent template actually reads (`USE_DATAROBOT_LLM_GATEWAY=0`, `OPENAI_API_BASE`, `OPENAI_API_KEY`, and an `openai/`-prefixed `LLM_DEFAULT_MODEL`) instead of the unread `EXTERNAL_LLM_*` keys. The external path now stops after writing `.env` for local `task dev`, skipping the DataRobot-only `dr dotenv setup` and Pulumi deploy.
+
 ## [1.8.0] - 2026-09-02
 
 ### Added
--- a/skills/datarobot-agent-assist/agent-assist-build/SKILL.md
+++ b/skills/datarobot-agent-assist/agent-assist-build/SKILL.md
@@ -56,7 +56,7 @@ Run in order before proceeding:
 3. **DataRobot CLI** — run `dr --version` and `dr auth check`. If either fails, invoke the `datarobot-setup` skill before continuing. Do not print manual install instructions.
 4. **Codespace** — run `python <skill_scripts_dir>/check_codespace.py` (no-op outside a Codespace). On non-zero exit, relay its message and stop; otherwise relay any exposed-ports warning it prints.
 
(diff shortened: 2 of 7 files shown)


chasdr:
folding this into the combined agent-assist litellm change built off victor's branch (external + litellm + disable-gateway/deployed in one pass, litellm on DATAROBOT_LITELLM_*). the env-contract fix here carries over as the .env writer there: OPENAI_API_BASE/OPENAI_API_KEY + an openai/-prefixed LLM_DEFAULT_MODEL + a generated SESSION_SECRET_KEY + USE_DATAROBOT_LLM_GATEWAY=0, verified end to end against the litellm proxy. closing so we dont have two parallel rewrites of the same files.

PR #101, opened 2026-09-02
[CFX-7986] agent-assist: LiteLLM and external LLM via the OpenAI-compatible .env [DO NOT MERGE]

chasdr:
## Summary

The agent-assist build skill could list an external OpenAI-compatible LLM (shipped in 1.8.0), but the `.env` it wrote used `EXTERNAL_LLM_*` keys the generated app's template never reads. Pointing an agent at an OpenAI-compatible endpoint silently failed: the model was never reached and setup died at the deploy step. This adds LiteLLM as a first-class model source alongside gateway, deployed, and external, and writes the OpenAI-compatible `.env` the template actually consumes. External and litellm now stop after writing `.env` for a local `task dev` instead of running the DataRobot-only deploy.

## Notes for review

External and litellm share one `.env` writer. The model carries an `openai/` prefix because litellm routes to `OPENAI_API_BASE` only for the `openai` provider and strips one prefix on the wire, so the raw model reaches the proxy verbatim. The listing is read only through `dr llm-gateway list`, so deployed LLMs need `dr` 0.2.79+.

## Output

```
$ setup_template.py --llm-source external --llm-model local-ollama --target-dir .
# with AGENT_ASSIST_LLM_BASE_URL=http://localhost:4000/v1 and AGENT_ASSIST_LLM_API_KEY set
DATAROBOT_ENDPOINT=
DATAROBOT_API_TOKEN=
LLM_DEFAULT_MODEL="openai/local-ollama"
USE_DATAROBOT_LLM_GATEWAY="0"
OPENAI_API_BASE="http://localhost:4000/v1"
OPENAI_API_KEY="ext-key"
SESSION_SECRET_KEY="<generated>"

$ setup_template.py --llm-source litellm --llm-model gpt-4o --target-dir .
# with DATAROBOT_LITELLM_BASE_URL=http://litellm:4000/v1 and DATAROBOT_LITELLM_API_KEY set
DATAROBOT_ENDPOINT=
DATAROBOT_API_TOKEN=
LLM_DEFAULT_MODEL="openai/gpt-4o"
USE_DATAROBOT_LLM_GATEWAY="0"
OPENAI_API_BASE="http://litellm:4000/v1"
OPENAI_API_KEY="litellm-key"
SESSION_SECRET_KEY="<generated>"
```

## Technical Changes

- `setup_template.py`: OpenAI-compatible `.env` for external and litellm, an early return that skips `dr dotenv setup` and the Pulumi deploy, and source-aware model verification.
[...]

PR #103, opened 2026-09-10
[CFX-7986] agent-assist: write the external LLM env the template reads

chasdr:

## What's broken

The agent-assist skill's external LLM path (shipped in 1.7.0) writes `EXTERNAL_LLM_MODEL`, `EXTERNAL_LLM_API_KEY`, and `EXTERNAL_LLM_BASE_URL` to the generated app's `.env`. Nothing in the agent template or its shared LLM library (`datarobot-genai`) reads those keys. So pointing an agent at an OpenAI-compatible endpoint (a local LiteLLM proxy on a box with no DataRobot LLM Gateway, say) silently fails: the model is never reached, and setup dies at the DataRobot deploy step on an unknown model.

## Why it's not just a key rename

The template resolves an external LLM from `USE_DATAROBOT_LLM_GATEWAY=0` plus `OPENAI_API_BASE`, `OPENAI_API_KEY`, and a model named `openai/<model>`, so litellm's OpenAI-compatible provider calls the base URL. Two more template behaviors fight a raw base URL: `dr dotenv setup` rewrites `.env` back onto the gateway path (dropping the OpenAI keys and injecting a prod endpoint), and the Pulumi deploy wraps every LLM in a governed DataRobot deployment. Neither fits an OpenAI-compatible endpoint.

## Fix

- `setup_template.py` external path writes what the template actually reads: `USE_DATAROBOT_LLM_GATEWAY=0`, `OPENAI_API_BASE`, `OPENAI_API_KEY`, an `openai/`-prefixed `LLM_DEFAULT_MODEL`, and a generated `SESSION_SECRET_KEY` (the FastAPI server requires it, `dr dotenv setup` normally generates it). Rejects a base URL or key carrying `.env`-breaking characters.
- Stops after writing `.env` for the external case, skipping `dr dotenv setup` and the Pulumi deploy. Runs locally with `task dev`.
- Docs (llm-selection, helper-scripts, agent-spec-schema, pre-coding-checklist, SKILL) describe the external source with the correct contract.
- Contract tests cover the OpenAI-compatible `.env`, the `openai/` prefix, credential rejection, and the skip-deploy early return.

## Scope and follow-ups

Local-run path only. Deploying an external-LLM agent is a separate follow-up. The external path also skips template framework materialization, so a non-default agentic framework (CrewAI, LlamaIndex, NAT) must be materialized via normal template setup first. The setup message says so.

This is the clean refile of the env-contract fix from closed #100 and #101. Env-contract only; the litellm-source rework from #101 is dropped.

## Verification

`task lint` clean, contract tests pass (7 new external cases). The env writer produces the OpenAI-compatible `.env` and runs no dotenv or deploy for the external case.

---

> [!NOTE]
> **Medium Risk**
> Changes agent project `.env` generation and setup flow for external LLMs; mistakes could break local dev or leak malformed secrets, though scope is limited to the external path and is covered by new contract tests.
> 
> **Overview**
> Fixes **datarobot-agent-assist** external OpenAI-compatible LLM setup so generated `.env` matches what the agent template and litellm actually consume.
> 
> **`setup_template.py`** no longer writes unused `EXTERNAL_LLM_*` keys. For `--llm-base-url` it sets `USE_DATAROBOT_LLM_GATEWAY=0`, `OPENAI_API_BASE`, `OPENAI_API_KEY`, an `openai/`-prefixed `LLM_DEFAULT_MODEL`, and `SESSION_SECRET_KEY`, and rejects base URL/API key characters that would break quoted `.env` lines. The external path **returns after writing `.env`**, skipping `dr dotenv setup` and Pulumi deploy so those steps do not rewrite the file or require a DataRobot LLM deployment.
> 
> **Docs and skill flow** add `llm_base_url` to the spec schema and document the third `external` model source (`AGENT_ASSIST_LLM_*`), including pre-coding `setup_template.py` invocation and that local dev uses `task dev` without dress rehearsal for external models.
> 
> **Contract tests** lock in the new `.env` shape, `openai/` model prefixing, credential validation, and the early-exit behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ad97680e8b877aede24aa69282a9c54763316e5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


andriykislitsyn (approved):

victorborshak (changes requested):
@chasdr  I spend sometime testing this, but I don't think it's working. I know, the implementation in the main branch is not quite a working one too, but this PR doesn't seem to be fixing it. I spend some time validating it manually and the issues I see are:
- the `task start-non-interactive` does not seem to be started at all (skipped?). I tried running `task start-non-interactive` or `dr start` manually after `.env` was created by agent assist but it fails somewhere at the pulumi stage.

but the bigger issue is:
- even thought the `OPENAI_API_BASE` & `OPENAI_API_KEY` are being saved in the .env file, looks like the template cannot use them alone to work properly. I think those variables can work only if `OPENAI_API_DEPLOYMENT_ID` & `OPENAI_API_VERSION` are provided too, which we obviously don't have. (I tried running `dr start` and selecting external model -> Azure with OpenAI - it requires `OPENAI_API_DEPLOYMENT_ID` + `OPENAI_API_VERSION` in addition to `OPENAI_API_BASE` & `OPENAI_API_KEY`.

So, I think it's either the template does not support what we're trying to do here, or we need to research more to actually make it work.

chasdr:
Parking this for the moment pending further exploration.

